Issue #10, opened 2025-12-04
[Issue] API Error: Cannot read properties of undefined (reading 'map')


KennyDizi:
I got this issue during testing the version v2.6.3

```
TypeError: fetch failed
    at node:internal/deps/undici/undici:15845:13
    at async OpenRouterHandler.handle (file:///Users/kennydizi/.nvm/versions/node/v24.11.1/lib/node_modules/claudish/dist/index.js:36041:22)
    at async dispatch (file:///Users/kennydizi/.nvm/versions/node/v24.11.1/lib/node_modules/claudish/dist/index.js:32970:17)
    at async cors2 (file:///Users/kennydizi/.nvm/versions/node/v24.11.1/lib/node_modules/claudish/dist/index.js:34613:5)
    at async dispatch (file:///Users/kennydizi/.nvm/versions/node/v24.11.1/lib/node_modules/claudish/dist/index.js:32970:17)
    at async file:///Users/kennydizi/.nvm/versions/node/v24.11.1/lib/node_modules/claudish/dist/index.js:33785:25
    at async responseViaResponseObject (file:///Users/kennydizi/.nvm/versions/node/v24.11.1/lib/node_modules/claudish/dist/index.js:34840:13)
    at async Server.<anonymous> (file:///Users/kennydizi/.nvm/versions/node/v24.11.1/lib/node_modules/claudish/dist/index.js:34973:14) {
  [cause]: SocketError: other side closed
      at TLSSocket.onHttpSocketEnd (node:internal/deps/undici/undici:7147:26)
      at TLSSocket.emit (node:events:520:35)
      at endReadableNT (node:internal/streams/readable:1701:12)
      at process.processTicksAndRejections (node:internal/process/task_queues:89:21) {
    code: 'UND_ERR_SOCKET',
    socket: {
      localAddress: '2001:ee0:4fb9:b7f0:dcb7:20e7:b016:6fed',
      localPort: 57109,
      remoteAddress: '2606:4700::6812:273',
  ⎿  API Error: Cannot read properties of undefined (reading 'map')
```

KennyDizi:
Model using `x-ai/grok-4.1-fast`

KennyDizi:
PR https://github.com/MadAppGang/claudish/pull/11

erudenko:
thank you for contribution

PR #11, opened 2025-12-05
Fix issue #10

KennyDizi:
Could you fix malformed parsing.

### DIFF
--- a/src/handlers/openrouter-handler.ts
+++ b/src/handlers/openrouter-handler.ts
@@ -104,17 +104,28 @@ export class OpenRouterHandler implements ModelHandler {
 
     await this.middlewareManager.beforeRequest({ modelId: target, messages, tools, stream: true });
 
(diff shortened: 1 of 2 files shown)


erudenko:
@claude Please review this PR and provide feedback on the implementation.

erudenko:
@claude Review this PR for code quality, potential bugs, and adherence to project patterns.

erudenko:
Thank you for the fix @KennyDizi! 

This PR was 67 commits behind main, so instead of merging with extensive conflicts, I've applied your fix directly to main in commit 7f00208.

**Changes applied:**
- Network error handling with try-catch around fetch
- `createStreamingErrorResponse` method for proper SSE error format
- Added `x-ai/grok-4.1-fast` model

The model selector performance optimization from your PR was not included as it would always fetch from the API even when cached models exist. The fallback to built-in config is a good idea but the implementation would need adjustment.

Issue #10 is now fixed. Thanks for the contribution!

PR #30, opened 2026-01-17
fix: add type coercion for tool arguments from local models

MidasMining:
## Summary

When using local models (like Seed-OSS via vLLM), tool calls like `TodoWrite` would fail with type mismatch errors:
```
The parameter todos type is expected as array but provided as string
```

This caused an infinite retry loop where the model kept trying to "fix" the format but kept getting rejected.

## Changes

Added a `coerceArgumentTypes()` function to `tool-call-recovery.ts` that:
- Parses strings as JSON when arrays/objects are expected
- Converts array-of-strings to array-of-objects for TodoWrite specifically
- Handles type mismatches gracefully before validation fails

## Test plan

- [x] Tested with Seed-OSS-36B-Instruct-AWQ via vLLM
- [x] Verified TodoWrite now works without infinite retry loops
- [x] Confirmed coercion only happens when type mismatch is detected

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/core/src/handlers/shared/tool-call-recovery.ts
+++ b/packages/core/src/handlers/shared/tool-call-recovery.ts
@@ -553,6 +553,96 @@ Format your tool calls as valid JSON with all required fields populated.
 }
 
 /**
+ * Coerce argument types to match the schema
+ * Handles cases where models output strings instead of arrays/objects
+ */
+export function coerceArgumentTypes(
+  args: Record<string, any>,
+  schema: ToolSchema
+): { args: Record<string, any>; coerced: boolean } {
+  if (!schema?.input_schema?.properties) {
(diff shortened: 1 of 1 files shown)


erudenko:
@claude Please review this PR and provide feedback on the implementation.

erudenko:
@claude Review this PR for code quality, potential bugs, and adherence to project patterns.

erudenko:
Thanks @MidasMining for this contribution! 🙏

This is a great fix for local model compatibility. The type coercion approach fits nicely with the existing tool-call-recovery module, and the TodoWrite-specific handling is pragmatic since that's the most common failure case.

The implementation is clean with good logging for debugging. Merging!

PR #37, opened 2026-01-24
fix: package.json scripts

svallory:
The scripts in `package.json` had the arguments in the wrong place


erudenko:
Thnak you for contribution

PR #38, opened 2026-01-25
Fix: API Key precedence and provider resolution

### LINKED ISSUE #35: Can't start without OPENROUTER_API_KEY
I run 
```
VERTEX_API_KEY=AIzaSyA82TqaXDNQEAkJCFtHfD0SDGsfw claudish --model v/gemini-2.5-flash "task"
```

but it returns:
```
Error: OPENROUTER_API_KEY environment variable is required
Get your API key from: https://openrouter.ai/keys

Set it now:
  export OPENROUTER_API_KEY='sk-or-v1-...'
```

I want to use direct API, why should I define openrouter api key?

svallory:
> [!IMPORTANT]
> This was implemented by Claude, please wait until I had a chance to review and test it before merging.

Fixes #35 - Users can now use provider-specific API keys without being incorrectly prompted for `OPENROUTER_API_KEY`.

## Problem

When running `claudish --model g/gemini-2.0-flash "task"` with `GEMINI_API_KEY` set, users got:
```
Error: GEMINI_API_KEY environment variable is required for model "g/gemini-2.0-flash"
```

Even though `GEMINI_API_KEY` was set!

### Root Cause

Two separate code paths handled provider resolution:

1. **Validation layer** (`provider-resolver.ts`) - Had fallback chain logic (provider → OpenRouter → Vertex)
2. **Routing layer** (`proxy-server.ts`) - Called `resolveRemoteProvider()` directly and threw if the provider's key was missing—**ignored fallback logic**

The routing layer was validating API keys AGAIN and throwing errors, completely bypassing the centralized fallback logic.

## Solution

Modified `getRemoteProviderHandler()` in `proxy-server.ts` to use `resolveModelProvider()` from `provider-resolver.ts` as the single source of truth for routing decisions.

### Key Changes

**Before:**
```typescript
const getRemoteProviderHandler = (targetModel: string): ModelHandler | null => {
  const resolved = resolveRemoteProvider(targetModel);

  // Validate API key
  const apiKeyError = validateRemoteProviderApiKey(resolved.provider);
  if (apiKeyError) {
    throw new Error(apiKeyError); // ❌ No fallback, just throws!
  }
  // ...
};
```

**After:**
```typescript
const getRemoteProviderHandler = (targetModel: string): ModelHandler | null => {
  // Use centralized resolver with fallback logic
  const resolution = resolveModelProvider(targetModel);

  // If resolver says use OpenRouter (including fallback cases), return null
  if (resolution.category === "openrouter") {
    return null; // Let OpenRouter handler take over
  }

  // Only create handler if direct-api AND key is available
  if (resolution.category === "direct-api" && resolution.apiKeyAvailable) {
    // ...create handler...
    return handler;
  }

  // Fall through to OpenRouter or return null
  return null;
};
```

### What This Achieves

1. **Single source of truth**: Routing layer now respects `provider-resolver.ts` decisions
2. **Fallback chain works**: provider → OpenRouter → Vertex (per README)
3. **Silent fallback**: When provider key missing but OpenRouter available, silently routes to OpenRouter
4. **No duplicate validation**: Removes the secondary validation that was throwing errors

## Changes

**Modified Files** (both `src/` and `packages/core/src/`):
- `proxy-server.ts`: Use `resolveModelProvider()` for routing decisions
  - Import `resolveModelProvider` from `provider-resolver.ts`
  - Check `resolution.category` and `resolution.apiKeyAvailable` before creating handlers
  - Return `null` to fall through to OpenRouter when appropriate
  - Added TODO noting manual sync issue between `src/` and `packages/core/src/`

**NOT Included** (saved for separate PR):
- Gemini error handling improvements
- Free tier quota messages
- Terminal quota detection

## Test Cases Now Working

```bash
# Direct API with provider key - Original Issue #35
GEMINI_API_KEY=xxx claudish --model g/gemini-2.0-flash "test"
# ✅ Should work without OPENROUTER_API_KEY

# Automatic fallback to OpenRouter when provider key unavailable
OPENROUTER_API_KEY=xxx claudish --model g/gemini-2.0-flash "test"
# ✅ Falls back to OpenRouter silently

# Automatic fallback chain: provider → OpenRouter → Vertex
VERTEX_API_KEY=xxx claudish --model g/gemini-2.0-flash "test"
# ✅ Falls back to Vertex when no Gemini or OpenRouter keys

# Local models - no key required
claudish --model ollama/llama3.2 "test"
# ✅ Works without any API key

# Both keys available - prefers provider-specific
GEMINI_API_KEY=xxx OPENROUTER_API_KEY=yyy claudish --model g/gemini-2.0-flash "test"
# ✅ Uses GEMINI_API_KEY directly (not OpenRouter)
```

## Architecture Improvement

This fix establishes `provider-resolver.ts` as the **single source of truth** for provider resolution logic, which:
- ✅ Reduces code duplication
- ✅ Makes fallback behavior consistent across all code paths
- ✅ Centralizes API key validation logic
- ✅ Makes future changes easier (one place to update)

## Technical Debt Noted

Added TODO comment noting that `/src/` and `packages/core/src/` are manually synced duplicates. This should be addressed in a future refactoring to establish proper monorepo structure.

### DIFF
--- a/packages/core/src/providers/provider-resolver.ts
+++ b/packages/core/src/providers/provider-resolver.ts
@@ -134,6 +134,22 @@ const LOCAL_PREFIXES = [
 ];
 
 /**
+ * Display names for providers (for proper capitalization)
+ */
+const PROVIDER_DISPLAY_NAMES: Record<string, string> = {
+  gemini: "Gemini",
+  "gemini-codeassist": "Gemini Code Assist",
+  vertex: "Vertex AI",
+  openai: "OpenAI",
+  openrouter: "OpenRouter",
+  minimax: "MiniMax",
+  kimi: "Kimi",
+  glm: "GLM",
+  ollamacloud: "OllamaCloud",
+  "opencode-zen": "OpenCode Zen",
+};
+
+/**
  * Check if any of the API keys (including aliases) are available
  */
 function isApiKeyAvailable(info: ApiKeyInfo): boolean {
@@ -162,6 +178,11 @@ function isApiKeyAvailable(info: ApiKeyInfo): boolean {
  *
  * This is THE single source of truth for provider resolution.
  * All code paths should call this function instead of implementing their own logic.
+ *
+ * Per README Model Routing table:
+ * 1. Try provider-specific API (if key available)
+ * 2. Fall back to OpenRouter (if OPENROUTER_API_KEY available)
+ * 3. Fall back to Vertex AI (if VERTEX_API_KEY or VERTEX_PROJECT available)
  *
  * @param modelId - The model ID to resolve (can be undefined for default behavior)
  * @returns Complete provider resolution including API key requirements
@@ -212,12 +233,12 @@ export function resolveModelProvider(modelId: string | undefined): ProviderResol
     };
   }
 
-  // 2. Check for direct API providers (g/, gemini/, v/, vertex/, oai/, mmax/, etc.)
+  // 2. Check for direct API providers with explicit routing prefixes
   const remoteResolved = resolveRemoteProvider(modelId);
   if (remoteResolved) {
     const provider = remoteResolved.provider;
 
-    // Special case: OpenRouter models with or/ prefix still go through OpenRouter
+    // Explicit OpenRouter prefix (or/) - always use OpenRouter
     if (provider.name === "openrouter") {
       const info = API_KEY_INFO.openrouter;
       return {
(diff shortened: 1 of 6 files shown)


svallory:
@claude

erudenko:
Hey @svallory! 🎉

Thank you so much for this contribution - really appreciate you taking the time to dig into the API key validation architecture!

Your centralized `ProviderResolver` approach is exactly right. I made a few small tweaks before merging:
- Reverted `google/` and `openai/` prefixes to route to OpenRouter (to avoid breaking existing workflows)
- Removed unused imports
- Added 41 unit tests for the provider resolution logic

Regarding your questions about the architecture:

> calling APIs directly instead of using a lib like ai sdk

Good observation! We handle things manually for a few reasons:
1. **Streaming compatibility** - Need precise control over SSE handling for Claude Code's streaming format
2. **Minimal dependencies** - Keeps the tool lightweight and fast to install
3. **Provider-specific quirks** - Each provider has subtle differences we need to handle

But I'm absolutely open to improvements! If you see ways to simplify things while keeping it robust, PRs are very welcome. The code duplication between `src/` and `packages/core/src/` is definitely something that should be cleaned up.

Thanks again for the solid foundation here - the ProviderResolver pattern makes future changes much easier! 🙏

svallory:
Hi, @erudenko! I'm glad you liked it!

> * Reverted `google/` and `openai/` prefixes to route to OpenRouter (to avoid breaking existing workflows)

I understand wanting to keep backwards compatibility, but this is very confusing to me. Specially when there's a specific "or/" prefix.

My question then is: how do I route to google? My whole goal was to make it possible to use my Google AI Studio key which links to my subscription.

Also, the behavior was to route to google when a GEMINI_API_KEY was present and fallback to OpenRouter when one wasn't. if you decide to keep the behavior of having "google/" map exclusively to OR, I suggest, at least, making this VERY clear in the README.

My honest take on how prefixes work now is that it is very hard to understand what Claudish will do, and that makes things really hard for newcomers.

erudenko:
Hey @svallory! 

Yes, absolutely agree that model routing has become super confusing. We initially had only OpenRouter models, but then started adding more providers and now it's a mess. 

I support your idea and think we need to explicitly indicate the provider now with less magic behind it - only skipping the provider for native API model usage.

**Good news:** We just released **v4.0.0** which addresses exactly this! 🎉

The new syntax is `provider@model[:concurrency]`:

```bash
# Explicit provider routing (new)
claudish --model google@gemini-2.0-flash "task"
claudish --model openrouter@google/gemini-pro "task"  
claudish --model oai@gpt-4o "task"

# Native auto-detection (no prefix needed for known models)
claudish --model gemini-2.0-flash "task"    # → Google API
claudish --model gpt-4o "task"               # → OpenAI API
claudish --model llama-3.1-70b "task"        # → OllamaCloud

# Unknown vendors now REQUIRE explicit openrouter@
claudish --model openrouter@deepseek/deepseek-r1 "task"
```

The old `g/`, `oai/` prefixes still work but show deprecation warnings.

Thanks again for pushing on this - your feedback helped shape the direction! 🙏

svallory:
> Thanks again for pushing on this - your feedback helped shape the direction! 🙏

You are very welcome, sir! Thank you for addressing this so quickly.

I'll continue to contribute as the need comes. I don't have much time to spare for now, but when I do and I'm more familiarized with the code I can help tackle bug reports and review PRs

PR #61, opened 2026-02-16
feat: Multi-Key API Key Rotation with Automatic Failover (KeyPool)

### LINKED ISSUE #60: feat: Multi-Key API Key Rotation with Automatic Failover (KeyPool)
## Summary

This feature introduces **KeyPool**, a transparent multi-key rotation and failover system for all API providers. Users can supply multiple API keys per provider via comma-separated environment variables, and Claudish will automatically distribute requests across keys with intelligent failover on errors.

```bash
# Before: single key, single point of failure
export GEMINI_API_KEY="key1"

# After: multiple keys, automatic rotation on errors
export GEMINI_API_KEY="key1,key2,key3"
```

No CLI changes, no configuration flags — just add more keys to the same environment variable.

## Motivation

API rate limits (HTTP 429) are a common friction point when using LLM providers heavily. When a key is rate-limited, the user must wait or manually switch keys. Similarly, expired or revoked keys cause hard failures with no recovery path. Multi-key rotation addresses this:

- **Rate limit resilience**: When one key hits 429, the next key is tried immediately
- **Key lifecycle management**: Expired or invalid keys are skipped automatically
- **Zero-downtime key rotation**: Add a new key to the list, remove old ones later
- **Team and CI use cases**: Distribute load across multiple API keys to stay within per-key quotas

### KeyPool Class

`KeyPool` is the core abstraction, located at `handlers/shared/key-pool.ts`:

- **Initialization**: Parses a comma-separated key string, trims whitespace, filters empty entries
- **Round-robin rotation**: Keys are tried in order; the index sticks on success and advances on failure
- **`executeWithFailover(fetchFn)`**: The main entry point — wraps a fetch function with automatic key rotation:
  - On **rotatable HTTP errors** (401, 402, 403, 408, 429, 500–504): rotates to the next key and retries
  - On **body-detected invalid keys** (non-rotatable status codes like 400): inspects the response body for provider-specific invalid key patterns before giving up
  - On **network errors** (thrown exceptions): rotates and retries
  - On **non-retryable errors** (e.g., 400 validation errors): returns immediately without wasting remaining keys
  - On **all keys exhausted**: returns the last failed response (preserving body for error details) or re-throws the last network error
- **Resource safety**: Drains previous response bodies to prevent connection and memory leaks, while preserving the last response for the caller

### Response Body-Based Invalid Key Detection

Some providers return non-standard status codes for invalid API keys. For example, Gemini returns HTTP 400 (not normally retryable) with `reason: "API_KEY_INVALID"` in the response body. The `isInvalidApiKeyResponse()` method clones the response and inspects the body for known patterns:

| Provider | Status | Detection Pattern |
|---|---|---|
| Gemini | 400 | `error.details[].reason === "API_KEY_INVALID"` |
| OpenAI | 401 | `error.code === "invalid_api_key"` |
| Anthropic | 401 | `error.type === "authentication_error"` |
| Generic | varies | Message contains `"API key not valid"`, `"invalid api key"`, `"invalid credentials"`, `"incorrect api key"` |

This provides defense-in-depth: even if a provider changes their error status code, the body-level check catches it. Non-JSON responses gracefully return `false` (no false positives).

### Handler Integration

All 7 provider handlers have been updated to use KeyPool:

| Handler | Provider | Key Env Var |
|---|---|---|
| `base-gemini-handler.ts` | Gemini | `GEMINI_API_KEY` |
| `openai-handler.ts` | OpenAI | `OPENAI_API_KEY` |
| `anthropic-compat-handler.ts` | Anthropic, MiniMax, Kimi | Provider-specific |
| `openrouter-handler.ts` | OpenRouter | `OPENROUTER_API_KEY` |
| `ollamacloud-handler.ts` | OllamaCloud | `OLLAMA_API_KEY` |
| `litellm-handler.ts` | LiteLLM | `LITELLM_API_KEY` |
| `remote-provider-handler.ts` | Generic remote providers | Registry-defined |

Each handler wraps its `fetch()` call inside `keyPool.executeWithFailover()`, with the API key injected per-attempt rather than once at construction time.

### Backward Compatibility

- **Single key**: KeyPool works transparently with a single key (no rotation, no extra logging)
- **No key**: Handlers that don't require keys continue to work unchanged
- **Logging**: Multi-key rotation is only logged when more than one key is configured, keeping single-key output clean

### Unit Tests (`tests/key-pool.test.ts` — 1,990 lines)

- **Core KeyPool**: Initialization (empty, single, multi-key, whitespace handling), round-robin rotation, index advancement, reset
- **executeWithFailover**: Success on first try, failover through keys, all-keys-exhausted (both HTTP and network errors), non-retryable status passthrough, response body draining, mixed error types
- **Status code handling**: Every code in `ROTATABLE_STATUS_CODES` verified, non-rotatable codes confirmed as passthrough
- **Invalid key detection**: Every detection pattern (Gemini, OpenAI, Anthropic, generic messages), negative cases (normal 400, non-JSON body), body preservation via clone, all-keys-exhausted with body detection
- **Concurrency resilience**: Concurrent `executeWithFailover` calls, large key pools (10+ keys), key exhaustion ordering

### Live Provider Canary Tests (5 tests)

Real HTTP requests to Gemini, OpenAI, Anthropic, OpenRouter, and OllamaCloud with deliberately invalid API keys. Each response is piped through `isInvalidApiKeyResponse()` to verify the detection logic matches what the provider actually returns. **If any provider changes their error format, the corresponding canary test fails** — alerting maintainers to update the detection logic. This eliminates the risk of silent regression from provider API contract changes.

## Files Changed (37)

- `packages/cli/src/handlers/shared/key-pool.ts` — new `KeyPool` class
- `packages/cli/src/handlers/shared/gemini-retry.ts` — Gemini-specific retry integration
- `packages/cli/src/handlers/shared/remote-provider-handler.ts` — generic remote provider KeyPool integration
- `packages/cli/src/handlers/*.ts` — all 7 handlers updated to use KeyPool
- `tests/key-pool.test.ts` — 116 tests (1,990 lines)
- `tests/gemini-retry.test.ts` — Gemini retry tests
- Synced to `packages/core/` and `src/`

ariel42:
Closes #60

## Overview

Introduces **KeyPool** — a transparent multi-key rotation and failover system. Users supply multiple comma-separated API keys per provider, and Claudish automatically distributes requests across them with intelligent failover on rate limits, authentication errors, and transient failures.

```bash
export GEMINI_API_KEY="key1,key2,key3"
export OPENAI_API_KEY="sk-abc, sk-def, sk-ghi"
```

Zero configuration — works with any existing single-key setup, no CLI changes needed.

### Failover Strategy

`executeWithFailover(fetchFn)` wraps each handler's fetch call:
- **Rotatable errors** (401, 402, 403, 408, 429, 500–504): rotate to next key, retry immediately
- **Body-detected invalid keys**: inspects response body for provider-specific patterns (e.g., Gemini returns 400 + `API_KEY_INVALID` — not normally retryable, but the body reveals it's a key issue)
- **Network errors** (thrown exceptions): rotate and retry
- **Non-retryable errors** (e.g., 400 validation): return immediately without wasting remaining keys
- **All keys exhausted**: return last response (preserving body for error details) or re-throw last error

### Response Body-Based Invalid Key Detection

`isInvalidApiKeyResponse()` clones the response and checks for known patterns:

| Provider | Status | Pattern |
|---|---|---|
| Gemini | 400 | `error.details[].reason === "API_KEY_INVALID"` |
| OpenAI | 401 | `error.code === "invalid_api_key"` |
| Anthropic | 401 | `error.type === "authentication_error"` |
| Generic | varies | Message-based matching (`"API key not valid"`, `"invalid credentials"`, etc.) |

Non-JSON bodies gracefully return `false` — zero risk of false positives.

### Resource Safety

Previous response bodies are drained between attempts to prevent connection and memory leaks. The last response body is preserved for the caller to read error details (quota info, retry-after hints, etc.).

## Handler Changes

All 7 provider handlers updated to use KeyPool:
- `base-gemini-handler.ts` / `gemini-handler.ts`
- `openai-handler.ts`
- `anthropic-compat-handler.ts`
- `openrouter-handler.ts`
- `ollamacloud-handler.ts`
- `litellm-handler.ts`
- `remote-provider-handler.ts`

Each handler injects the API key per-attempt via the failover callback rather than once at construction time.

## Test Coverage (116 tests, 1,990 lines)

- **Core KeyPool unit tests**: Initialization, round-robin rotation, index management, reset
- **executeWithFailover integration**: Success/failure paths, mixed error types, concurrent calls, large pools (10+ keys), non-retryable passthrough
- **Status code coverage**: Every rotatable code verified, non-rotatable codes confirmed as passthrough
- **Invalid key detection**: All provider-specific patterns, negative cases, body preservation
- **5 live canary tests**: Real API calls to Gemini, OpenAI, Anthropic, OpenRouter, and OllamaCloud with invalid keys — responses piped through `isInvalidApiKeyResponse()` to catch provider API contract changes

## Stats

37 files changed, 4,337 insertions, 321 deletions


erudenko:
The KeyPool core is well-designed. Round-robin with failover, body-draining between attempts, preserving the last response for error details, and the `ROTATABLE_STATUS_CODES` set is well-chosen. The body-based invalid key detection for Gemini's unusual 400 response is smart.

But the integration into handlers needs rework:

Every handler has the same `if (keyPool.keyCount() > 1) { ... failover ... } else { ... existing ... }` copy-pasted. `executeWithFailover` already handles single-key correctly (one loop iteration), so the branching is unnecessary. It doubles the code in each handler for zero benefit. Just always use `executeWithFailover`.

The changes are tripled across `src/`, `packages/cli/src/`, and `packages/core/src/`. Only `src/` is the source of truth. This inflates the PR from ~1,400 meaningful lines to 4,247.

In `base-gemini-handler.ts`, multi-key mode uses `fetchWithRetry` with `maxRetries: 1, skipRetryOn429: true` while single-key gets `maxRetries: 5`. So multi-key users get significantly less resilience on transient errors. The two retry strategies should compose, not replace each other.

The Kimi OAuth fallback is inside the single-key else branch in `anthropic-compat-handler.ts`. Multi-key users of Kimi silently lose OAuth support. That's a feature regression.

Also note that PR #70 (provider refactor) rewrites most of the same files. If that merges first, this PR will need a full rewrite against the new architecture. Might be worth waiting to see how #70 lands.

The feature itself is valuable though. Happy to help figure out the right integration approach.

erudenko:
Putting this on hold until PR #70 (provider refactor) lands. That PR rewrites every handler file this touches, so integrating KeyPool against the current architecture would just be throwaway work.

The KeyPool class itself is solid and we'll use it. Once #70 gives us the new 3-layer architecture, the integration will be cleaner since there's one handler (`ProviderHandler`) instead of six separate ones with copy-pasted fetch logic.

I'll ping you when #70 is in and we're ready to wire KeyPool into the new transport layer.

erudenko:
Closing this, and the timing on it was rough. You opened it on 16 February, and `fdcadd5` (v5.0.0, composable handler architecture) deleted every file it touches on 19 February. Three days. `packages/core`, the top-level `src`, `tests`, and all seven per-provider handlers are gone.

The feature is still missing though. No KeyPool anywhere on main. So this isn't a "we don't want it" close.

`key-pool.ts` itself is good. Self-contained, drains previous bodies so they don't leak, keeps the last response around for error detail. The rotatable status set and the provider pattern table in `isInvalidApiKeyResponse` are the parts I'd keep more or less verbatim.

The reason it can't just be re-pointed at the new files is that key handling centralised while it sat. `api-key-credential.ts` resolves one key string, memoizes it, and writes it into `process.env` so spawned children inherit it. `getRequestAuth()` bakes it into an Authorization header, proxy-server extracts it back out, and the handler gets built once and cached. So a comma-separated value today goes out as a single bearer token and 401s every time.

The better news is the rewrite is much smaller than what you originally wrote. There's exactly one outbound fetch pair in the whole pipeline now, `composed-handler.ts:521` and its 401-retry twin at `:624`, both fed by `this.provider.getHeaders()`. So a multi-key `ApiKeyCredential` whose `getRequestAuth()` takes an attempt index gets you the entire feature, and it composes with 1Password, pre-hydration and route pinning rather than fighting them. Your seven-handler integration was about 2000 lines of the diff, and centralising deleted the need for all of it.

One thing to carry across: `credentials.isAvailable()` would report available when only the first key in a pool is any good. Worth handling in the new shape.

If you want to take that on I'll review it fast. If not I'll get to it, with credit on the commit. Either way, sorry it sat for six months.

erudenko:
Actually closing it this time. I wrote "Closing this" on 9 August and then never hit the button, so this has been sitting open under a comment saying it was done. That's on me.

Everything in that comment still stands: `key-pool.ts` is good work, the feature is still missing from main, and the reason it can't be re-pointed at the new files is that key handling centralised into `api-key-credential.ts` while this sat.

Nothing has changed since to make a rebase easier, so I'd rather stop it pretending to be live.

PR #63, opened 2026-03-02
fix: add ToolSearch to tool-call-recovery inference

mtparet:
Adds ToolSearch to the knownTools list for natural language extraction, inference logic for missing max_results (defaults to 5) and query parameters, and system prompt guidance for local models.

Fixes incomplete ToolSearch tool calls from models like MiniMax-M2.5 (or qwen 3.5) that omit the required max_results parameter.

### DIFF
--- a/packages/cli/src/handlers/shared/tool-call-recovery.ts
+++ b/packages/cli/src/handlers/shared/tool-call-recovery.ts
@@ -179,6 +179,7 @@ export function extractToolCallsFromText(text: string): ExtractedToolCall[] {
     "Glob",
     "WebFetch",
     "WebSearch",
+    "ToolSearch",
   ];
   const nlPatterns = [
     // "I'll use the X tool with param=value" - ends with period, colon, newline, or end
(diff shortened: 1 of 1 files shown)


erudenko:
The change follows the existing tool-call-recovery patterns exactly, and defaulting `max_results` to 5 matches the schema default. Good addition.

Same file path note as other PRs: the edit is in `packages/cli/src/` but should be in `src/` since that's the source of truth. The build copies from there.

One minor thing: the fallback `inferred.query = inferred.search || inferred.keyword || inferred.tool || ""` sets query to an empty string when everything is undefined. Might be better to leave it unset so the tool validation surfaces the actual error instead of silently sending an empty query. Not a big deal either way.

Can you move the change to `src/handlers/shared/tool-call-recovery.ts`?

erudenko:
Merged. ToolSearch was the one tool missing from the recovery list and MiniMax-M2.5 hits this exact failure mode. The inference fallback chain for query is a nice touch. Thanks Matthieu.

PR #68, opened 2026-03-12
fix: upgrade MCP SDK to fix Zod 4 tool schema serialization

MayCXC:
## Summary

MCP tool schemas (`--mcp` mode) serialize as empty `{"type":"object","properties":{}}` because the bundle contains two copies of Zod.

MCP SDK 1.22.0 declares `zod: ^3.23.8` (Zod 3 only), but claudish uses Zod 4. Bun resolves both separately and inlines them as two distinct exports in `dist/index.js`. The SDK then wraps Zod 4 raw shapes with Zod 3's `z.object()`, and `zodToJsonSchema` reads `_def.typeName` (which doesn't exist in Zod 4), producing empty output.

## Fix

- Bump `@modelcontextprotocol/sdk` from `^1.22.0` to `^1.27.0` in both root and `packages/cli`
- SDK 1.27.0 declares `zod: ^3.25 || ^4.0`, so Bun resolves a single Zod 4 for the whole tree
- The SDK also ships `zod-json-schema-compat.js` which uses Zod 4's native `toJSONSchema()` for serialization
- Move bundled deps to `devDependencies` since `bun build` inlines everything into `dist/index.js`

Related: modelcontextprotocol/typescript-sdk#1643

## Test plan

- [x] `bun install && bun run build` produces bundle with single Zod copy
- [x] `claudish --mcp` returns correct `inputSchema` with properties, types, and required fields for all 4 tools
- [x] Installed via `npm install -g --install-links` from built package, verified in Claude Code session

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -306,41 +306,38 @@ jobs:
       - name: Update Homebrew tap
         uses: actions/checkout@v4
         with:
-          repository: MadAppGang/homebrew-claudish
+          repository: MadAppGang/homebrew-tap
           token: ${{ secrets.HOMEBREW_TAP_TOKEN }}
           path: tap
 
       - name: Update formula
         run: |
-          cat > tap/Casks/claudish.rb << 'EOF'
-          cask "claudish" do
+          mkdir -p tap/Formula
+          cat > tap/Formula/claudish.rb << EOF
+          class Claudish < Formula
+            desc "Multi-model AI CLI - run Claude Code with any model"
+            homepage "https://github.com/MadAppGang/claudish"
             version "${{ steps.release.outputs.version }}"
+            license "MIT"
 
             on_arm do
-              url "https://github.com/MadAppGang/claudish/releases/download/v#{version}/claudish-darwin-arm64"
+              url "https://github.com/MadAppGang/claudish/releases/download/v${{ steps.release.outputs.version }}/claudish-darwin-arm64"
               sha256 "${{ steps.release.outputs.arm64_sha }}"
             end
 
             on_intel do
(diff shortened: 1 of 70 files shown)


erudenko:
Good investigation. The dual-Zod `_def.typeName` mismatch causing empty `{}` schemas is subtle, I wouldn't have caught it from the symptoms alone.

Moving deps to devDependencies makes sense since `bun build` inlines everything into `dist/index.js` anyway. The empty `dependencies: {}` looks a bit odd, could just drop the key entirely, but not a blocker.

Two things I'd like to check before merging:

1. The new SDK (1.27.0) pulls in its own `hono` dependency. We already use Hono for the proxy server so I want to verify there's no conflict or bundle size explosion from having two copies.

2. The diff includes a version bump from `5.10.0` to `5.13.1` in `packages/cli/package.json` which is unrelated to the Zod fix. Can you rebase on current main so that change doesn't sneak in as part of this PR?

Would also be great to have a test that asserts MCP tool schemas serialize with non-empty `properties`, so this doesn't regress. But I can add that separately after merge if you'd prefer.

erudenko:
Applied the SDK upgrade directly on main since rebasing would've been a headache with the version drift. Bundle size stayed at 4.1M, no Hono duplication issues. Thanks for tracking down the dual-Zod root cause, that was solid debugging.

erudenko:
Applied on main with credit. Tracking down the dual-Zod _def.typeName mismatch as the root cause of empty {} schemas is the kind of debugging that saves everyone hours. I've been staring at those empty tool schemas for a week and didn't connect it to the SDK's bundled Zod 3 conflicting with our Zod 4. Thanks Aaron.

PR #75, opened 2026-03-17
feat: add --pipe mode for automation + fix --json output capture

talmosko-code:
## Problem

Claudish has two issues that prevent programmatic/automation use:

### 1. `--json` output is broken

`claude-runner.ts` uses `stdio: "inherit"` in `spawn()`, so Claude Code's JSON output goes directly to the terminal instead of being captured by claudish. Result: JSON wrapper with empty `result` field.

**To reproduce:**
```bash
echo "hello" | npx claudish --model openrouter/healer-alpha --json
# Result: {"result": "", ...}  ← empty!
```

**Root cause:** Line 387 in `claude-runner.ts` — `stdio: "inherit"` means Claude Code's stdout is connected directly to the parent's stdout. Claudish never sees the output, so it generates JSON with an empty result.

### 2. No pipe/automation mode

Current options:
- **Interactive** (`claudish`): Shows TUI, requires PTY — unreliable for automation
- **Single-shot** (`claudish --model X "prompt"`): Exits after one prompt
- **`--stdin`** (`echo "prompt" | claudish --stdin`): Same as single-shot, exits after one prompt

What's missing: a persistent mode for sending multiple prompts programmatically while maintaining session context. Essential for bots, CI/CD pipelines, and automation frameworks.

### Fix 1: Capture stdout when `--json` is used
When `config.jsonOutput` is true, spawn with `stdio: ["inherit", "pipe", "inherit"]` to capture Claude Code's stdout before it reaches the terminal.

### Fix 2: New `--pipe` flag
Persistent stdin/stdout mode for automation:
- Spawns Claude Code with `-p --output-format stream-json --verbose`
- Relays stdin → Claude Code's stdin and Claude Code's stdout → stdout
- No TUI — clean stream I/O for programmatic use

```bash
# Pipe mode for automation:
echo "implement feature X" | claudish --pipe --model openrouter@provider/model

# JSON output (now fixed):
echo "hello" | claudish --json --model google@gemini-3-pro
```

## Changes
- `types.ts`: Add `pipe: boolean` config field
- `cli.ts`: Add `--pipe` flag parsing, exclude from interactive fallback, update help text and examples
- `claude-runner.ts`: Pipe mode (piped I/O with stdin/stdout relay) + JSON mode (capture stdout instead of inheriting)

## Why this matters

Claudish's tagline is "Run Claude Code with any AI model." Without programmatic access, it can only be used by humans at a terminal. This limits use in:
- **CI/CD pipelines** — automated code review
- **Chatbots and Discord bots** — programmatic responses
- **AI automation frameworks** — integration with tools like OpenClaw
- **Batch processing** — running multiple prompts sequentially
- **Testing and benchmarking** — programmatic model comparison

The pipe mode makes claudish a proper API-like interface for Claude Code, usable both interactively (TUI) and programmatically (pipe).

Happy to iterate on the implementation based on feedback!


erudenko:
Good problem identification on the JSON capture issue with \`stdio: "inherit"\`.

A few things I noticed:

In the JSON mode branch, \`capturedOutput += data.toString()\` accumulates all stdout but the variable is never read afterward. The fix pipes stdout through to the user, which is the same as before. The captured data goes nowhere.

The docs say pipe mode is "persistent stdin/stdout loop" where "each line on stdin is a prompt" but Claude Code's \`-p\` flag (which this adds) exits after one prompt. So \`--pipe\` is functionally the same as \`--stdin\` with JSON output, not a persistent loop.

The stdin relay has no backpressure handling. \`proc.stdin?.write(data)\` doesn't check the return value, so large inputs could lose data if Claude Code's buffer fills up. Fine for small prompts but worth noting.

\`--verbose\` is always added in pipe mode, which could be noisy for automation use cases where you only want the structured JSON.

Also if Claude Code exits unexpectedly, the stdin relay listeners are never cleaned up. Should handle \`proc.on("close")\`.

The concept is useful though. Both the JSON capture fix and the pipe mode have real use cases, they just need different implementations. Happy to discuss the right approach if you want to iterate on this.

erudenko:
Took a closer look. The use cases are real (machine-readable output for automation, stdin relay for scripting) but the implementation has a few gaps. Let me be specific so it's easy to fix.

**1. `capturedOutput` is accumulated but never used**

```typescript
// Line ~420 in the jsonOutput branch
capturedOutput += data.toString();
// Also write to real stdout so user sees progress
process.stdout.write(data);
```

`capturedOutput` grows with every chunk but nothing ever reads it. After the process exits, the variable is just garbage collected. If the intent was to re-emit structured JSON at the end, that code is missing. If you just need passthrough, drop `capturedOutput` entirely and only keep `process.stdout.write(data)`.

**2. Pipe mode isn't actually persistent**

The docs and help text say:
```
# Each line on stdin is a prompt; responses stream to stdout as JSON
```

But the implementation adds `-p` to Claude Code args, which is print mode. Claude Code in `-p` mode takes one prompt and exits. So `--pipe` behaves identically to `--stdin` with `--output-format stream-json`. There's no loop, no line-by-line prompt processing.

If you want actual persistent mode, you'd need to either:
- Spawn a new Claude Code process per line from stdin
- Or use Claude Code's conversation mode and relay messages through, which is closer to what the stdin relay code does but Claude Code doesn't support that via `-p`

If single-shot is the intended behavior, the docs should say that instead of "persistent stdin/stdout loop."

**3. No cleanup on process exit**

```typescript
process.stdin.on("data", (data: string) => {
  proc.stdin?.write(data);
});
```

If Claude Code exits (crash, timeout, ctrl-c from inside), these listeners stay attached to `process.stdin` forever. The Node process won't exit cleanly because stdin is still in flowing mode with `resume()`. Need something like:

```typescript
proc.on("close", () => {
  process.stdin.pause();
  process.stdin.removeAllListeners("data");
});
```

**4. No backpressure on stdin relay**

`proc.stdin?.write(data)` returns a boolean indicating if the buffer is full. If it returns false, you should pause `process.stdin` until `proc.stdin.on("drain")` fires. For small prompts this won't matter but piping a large file will lose data silently.

**5. `--verbose` forced in pipe mode**

```typescript
claudeArgs.push("--verbose");
```

Pipe mode is for automation. Automation tools parse structured output. `--verbose` adds human-readable noise that breaks parsers. This should be opt-in (user passes `-v` themselves), not forced.

**6. `--dangerouslyDisableSandbox` vs `--dangerously-skip-permissions`**

The pipe mode checks both `config.autoApprove` and `config.dangerous` but single-shot mode only handles `autoApprove`. If these flags matter for pipe mode they should work consistently across all modes, otherwise it looks like pipe mode has special safety bypass logic.

The flag parsing in `cli.ts` is clean though. `config.pipe = true` with `interactive = false` and the interactive-mode guard update are all correct.

Happy to help think through the persistent-mode design if that's what you're going for.

erudenko:
The need here is real. We don't have a multi-turn programmatic session. `--stdin` reads once and exits, one prompt one answer, no continuity. So I'm not closing this because the idea is wrong.

But the implementation doesn't do what the description says, and I'd rather explain than just decline.

The pipe branch pushes `-p --output-format stream-json --verbose`, but not `--input-format stream-json`. Without that, `claude -p` takes stdin as one prompt and exits. So "persistent stdin/stdout loop, multiple prompts maintaining session context" is a single turn, and every line after the first goes into a closed pipe.

`capturedOutput` is declared, appended to, and never read by anything in the diff. So the `--json` fix is a buffer copy with no consumer on the other end.

That half is also built on a diagnosis I don't think holds. We spawn with `stdio: "inherit"` and forward `--output-format json` straight to Claude Code, which writes its JSON to your stdout directly. Claudish doesn't build a wrapper on that path, so there's no empty result field to fill. The one place we do emit `{responses: ...}` is team-json mode, and team-orchestrator already spawns children with piped stdio and captures per-model output to files.

Two smaller ones. `proc.stdin?.write(data)` has no backpressure handling, so a large piped input drops silently on a full pipe. And the `--json` branch flips Claude Code's fd 1 from a TTY to a pipe, which matters because Claude Code decides print-versus-interactive from its own stdout. There's a long comment at `claude-runner.ts:1313` about exactly that.

There's also a naming problem to settle before any of this lands. `-p` is `--profile` in claudish but `--print` in Claude Code, and adding `--pipe` next to `--stdin` gives us three pipe-shaped surfaces. That needs a decision, not another flag.

The shape that would work is `-p --input-format stream-json --output-format stream-json` with a JSONL relay. Different PR, and I'd take it. Closing this one, but thanks for putting the time in.

erudenko:
Two halves here and they've aged differently. Closing, but read the second half before you write it off.

**The `--json` bug is gone.** It was real when you filed it, against an implementation that no longer exists. There's no claudish-built JSON wrapper anymore, so nothing can hand you an empty `result` field. `--json` now does one thing:

```ts
if (config.jsonOutput) {
  config.quiet = true; // JSON output mode is always quiet
}
```

Claudish stops talking and Claude Code's own output goes straight through. So `echo hello | claudish --model X --json` gives you Claude Code's JSON rather than a wrapper around it. Different shape from what you were expecting, but not empty.

**`--pipe` is the part still worth talking about.** Your `stdio: ["pipe","pipe","inherit"]` is genuinely different from what's there. `--stdin` and `-p` are single-shot: one prompt in, one answer, process exits. Yours holds a persistent loop, which is a real gap if you're driving claudish from another program across several turns.

I'm not merging it as-is because the branch is against a tree that's moved a lot (`claude-runner.ts` now picks stdio from a `ttyFd` it may or may not have, so the three-way branch lands differently), and because I'd want to know what happens on the edges: what ends a turn, what happens when the child dies mid-stream, whether a partial line can be split across reads.

If you still want this, open a fresh PR against main with just `--pipe` and I'll review it properly. Drop the `--json` part.

Worth saying: channel mode may already cover your use case. `create_session` / `send_input` / `get_output` over MCP is a persistent multi-turn session driven by another program, which sounds like the thing you were building toward. If that fits, `--pipe` may not need to exist. If it doesn't fit, that's useful for me to hear.

PR #79, opened 2026-03-19
refactor: replace DialectManager class with resolveModelDialect function

MayCXC:
## Summary

Replace the stateless `DialectManager` class with a pure `resolveModelDialect()` function. The class had no state: constructor allocated all dialects eagerly, `getAdapter()` did a linear scan, `needsTransformation()` called `getAdapter()` again. The function does the same work without the class ceremony.

Rebased onto v6.2.2.

- `DialectManager` class replaced by `resolveModelDialect(modelId): BaseAPIFormat`
- All callers updated: composed-handler, cli, local-adapter, openrouter-api-format, all test files
- CLAUDE.md updated
- 11 files changed, -54 lines net

## Test plan

- [x] `bun test` passes (149/149 across 3 test files)
- [x] All dialect selection tests pass (correct routing, false positive prevention)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -95,7 +95,7 @@ Each converter declares its stream format via `getStreamFormat()`.
 Translates model-specific dialect differences (context windows, thinking→reasoning_effort, vision rules).
 - **Interface**: `adapters/model-translator.ts`
 - **Implementations**: GLMAdapter, GrokAdapter, MiniMaxAdapter, DeepSeekAdapter, QwenAdapter, CodexAdapter
-- **Selection**: `AdapterManager` auto-selects based on model ID
+- **Selection**: `resolveModelDialect(modelId)` selects based on model ID
 
 ### Layer 3: ProviderTransport — HTTP transport
 Handles auth, endpoints, headers, rate limiting. Optionally overrides stream format for aggregators.
@@ -113,7 +113,7 @@ transport.overrideStreamFormat() ?? modelAdapter.getStreamFormat() ?? providerAd
 ```
 
 **Adding a new provider**: Add one entry to `PROVIDER_PROFILES` table in `providers/provider-profiles.ts`.
-**Adding a new model**: Create a ModelTranslator adapter, register in `adapters/adapter-manager.ts`.
+**Adding a new model**: Create a ModelDialect adapter, add it to the `DIALECTS` array in `adapters/dialect-manager.ts`.
 **Verifying wiring**: `claudish --probe <model>` shows the full adapter composition.
 
 ### Stream Parsers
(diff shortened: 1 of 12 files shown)


erudenko:
This one's still on target. `DialectManager` is unchanged in substance at `adapters/dialect-manager.ts:23`. Eager construction of 9 dialects in the constructor, `getAdapter()` doing a linear scan, `needsTransformation()` calling `getAdapter()` a second time. Exactly what you were removing.

We did the naming half independently since March. `AdapterManager` became `DialectManager`, and the files moved to `*-model-dialect.ts` and `*-api-format.ts`. So your rename hunks are already in. The class-to-function collapse is the part that's still open.

Two things moved under it. There's a second constructor param now, `wireFormat?: StreamFormat` at `dialect-manager.ts:39`, and dropping it would quietly regress the Qwen Plan multi-vendor endpoint behaviour. And it passes `this.bareModelName` rather than `targetModel`. So the signature wants to be `resolveModelDialect(modelId, wireFormat?)`.

I checked the equivalence question you'd otherwise have to argue: main's `getAdapter()` matches on `this.defaultAdapter.getModelId()`, and `getModelId()` returns the string unmodified, so your `d.shouldHandle(modelId)` is the same thing. That part holds.

Given the diff is five months old and two of the touched test files were renamed since, I think writing it fresh as a ~40 line commit beats fighting the rebase. That's what I'm planning, with credit to you. Say the word if you'd rather rebase it yourself, genuinely happy either way.

One question. Does anything outside `packages/cli` use the `BaseModelAdapter` and `AdapterManager` back-compat aliases in `adapters/index.ts`? You delete them and I haven't checked macos-bridge yet.

erudenko:
Landed in #202. Thank you for the patience on this one, it sat far longer than it should have.

I wrote it fresh rather than rebasing, which we'd discussed, and the diff is close to what you had: `resolveModelDialect(modelId, wireFormat?)`, candidates as factories so construction stops at the first match, `needsTransformation()` deleted rather than ported since nothing called it. Net minus sixteen lines. You're co-author on the commit.

Two notes on things that moved under the original diff, in case you're curious how it aged.

The class picked up a second constructor parameter, `wireFormat?: StreamFormat`. Dropping it would have quietly regressed Alibaba's Qwen Plan endpoint, where one Anthropic URL serves qwen3.x, glm-5.2 and deepseek-v4-* and each wants a different reasoning knob. So the signature became `resolveModelDialect(modelId, wireFormat?)` rather than taking the model alone.

And the equivalence you'd have had to argue does hold. The old `getAdapter()` matched on `this.defaultAdapter.getModelId()` rather than on the argument, and I checked: `BaseAPIFormat`'s constructor stores `modelId` unmodified and `getModelId()` returns it, so they're the same string. I noted it in the commit message as the one load-bearing claim in the diff, worth re-checking if either method ever changes.

Your read was right in March and it was still right in August. The class was doing a function's job.

PR #70, opened 2026-03-13
refactor: isolate provider definitions + custom providers

MayCXC:
## Summary

Adding a new provider to claudish currently requires editing 7+ constants across 3 files. Provider identity (name, shortcuts, API key, display name, transport type) is scattered across `model-parser.ts`, `remote-provider-registry.ts`, and `provider-resolver.ts`. The proxy server dispatches handlers via a 170-line if/else chain on provider name.

This refactor replaces all of that with a single `BUILTIN_PROVIDERS` array as the source of truth, a component selector that assembles the right transport + adapters from a definition, and a clean FormatAdapter/ModelAdapter separation. It also adds user-defined providers via `~/.claudish/config.json`.

Supersedes #69 (GitHub Models provider).

## New features

**GitHub Models provider** (`gh@`): Routes to `models.github.ai` with `GITHUB_MODELS_TOKEN`. Just one entry in `BUILTIN_PROVIDERS`.

**MCP direct provider support**: `run_prompt` routes through all providers (not just OpenRouter). `g@gemini-2.5-flash`, `gh@gpt-4o-mini`, `zen@gpt-5-nano`, etc. all work as MCP tool calls.

**User-defined providers**: `~/.claudish/config.json` with a `providers` map. Pick any existing transport type (`openai`, `anthropic`, `gemini`, `local`, etc.), set `baseUrl` and `apiKeyEnvVar`, and the new provider works everywhere: CLI, proxy, and MCP.

## Refactors

**Provider definitions** (`provider-definitions.ts`): Single `ProviderDefinition[]` array for all providers. Shortcuts, legacy prefixes, native model patterns, API key info, and display names are all derived from it. Replaces 6 scattered maps/constants.

**Provider component selector** (`provider-components.ts`): `selectProviderComponents(def, modelName, apiKey)` returns `{ transport, formatAdapter, modelAdapter }`. Three internal selectors dispatch by transport type (transport, format adapter) or model name (model adapter) using `parseModelSpec` for clean name matching. No `shouldHandle` polymorphism; all selection is inline in the selector. `ProviderComponents` contains only the three components; `tokenStrategy` and `summarizeTools` are passed directly by the caller to `ProviderHandler`. `TransportConfig` is assembled inside `selectTransport`, invisible to the caller.

[...]

PR #80, opened 2026-03-19
refactor: three-layer provider registry with centralized resolution

MayCXC:
## Summary

Centralize all provider-specific logic in `provider-profiles.ts`. Five resolvers, one `createHandlerForProvider` entry point. No provider branching in adapters, transports, stream parsers, composed-handler, or proxy-server. Rebased onto v6.2.2. Depends on #79.

### provider-profiles.ts (single source of truth)

Five resolvers + one composer:
- `resolveTransport(def, modelName, apiKey)` switches on `def.transport`
- `resolveAPIFormat(def, modelName)` switches on `def.transport`
- `resolveModelDialect(modelId)` returns `BaseModelDialect | null` (null = no quirks)
- `resolveMiddlewares(modelId)` selects middleware instances
- `resolveEffective(def, modelName, apiKey)` handles definition swaps (zen+minimax) and publicKeyFallback
- `createHandlerForProvider(def, modelName, apiKey, targetModel, port, opts)` composes all five

### Class hierarchy

**BaseAPIFormat** (wire format only): convertMessages, convertTools, buildPayload, getStreamFormat, parseResponse. Implementations: OpenAI, Anthropic, Gemini, Codex, Ollama, LiteLLM, OpenRouter.

**BaseModelDialect** (model quirks only): processTextContent, getContextWindow, supportsVision (default false), prepareRequest. Implementations: Grok (XML tool format injection), Gemini (reasoning filter, output format injection), DeepSeek, Qwen, MiniMax, GLM, Xiaomi.

**BaseTransport** owns RequestQueue with `onResponse()`/`shouldRetry()`/`calculateDelay()` as overridable methods. `ApiKeyTransport` (Bearer), `OAuthTransport` (token refresh). Transport metadata: tokenStrategy, getNonStreamingEndpoint (URL-parsed, strips query params), unwrapResponse, preserveVendorPrefix.

### Transport subclasses
- ZenTransport: extends OpenAI, public key fallback, GPT endpoint override
- KimiCodingTransport: extends Anthropic, OAuth fallback in getHeaders()
- OllamaTransport: extends Local, num_ctx injection, /api/tags health check, /api/show context window
- LMStudioTransport: extends Local, /v1/models context window detection

### Adapter subclasses
- GeminiModelDialect: extracted reasoning filter from GeminiAPIFormat
- LocalQwen/DeepSeek/Llama/MistralFormatAdapter: model-family sampling params and tool guidance

### Zen definition swap
- opencode-zen-minimax and opencode-zen-go-minimax definitions with transport: "anthropic"
- resolveEffective swaps definitions for minimax models
- Resolvers see the correct transport type, no model branching needed

### proxy-server.ts
Single `handlerCache` + `getHandler`. No transport or adapter imports. All construction via `createHandlerForProvider`.

### Removed
- `dialect-manager.ts`, `remote-provider-registry.ts`
- `shouldHandle` from all adapters, ModelDialect interface, middleware interface, MiddlewareManager filter
- `DialectManager` class, `ProfileContext`, `ProviderProfile` interface, `PROVIDER_PROFILES` table
- Standalone hook functions (geminiOnResponse, createOpenRouterHooks)
- `unwrapGeminiResponse` and `tokenStrategy` from ComposedHandlerOptions
- `hasNativeAnthropicMapping` from claude-runner
- Dead getter functions from provider-definitions
- Model-family branches from local-adapter, openai-messages, openrouter-api-format
- Gemini middleware check from composed-handler (middleware manager no longer filters)

### processTextContent chain
Stream parsers accept optional `modelDialect` parameter. Format adapter runs first, then dialect if format didn't transform. No proxy objects.

### provider-definitions.ts
Pre-computed maps at module load: PROVIDERS_BY_NAME, PROVIDER_SHORTCUTS, LEGACY_PREFIX_PATTERNS, NATIVE_MODEL_PATTERNS, API_KEY_INFO, PROVIDER_DISPLAY_NAMES. ProviderResolution gains def and apiKey fields.

## Test plan

- [x] 138/138 format-translation + provider-routing + glm-adapter tests pass
- [x] Full suite: 521 pass, 5 skip, 1 pre-existing fail (mtm binary)
- [x] Gemini MCP non-streaming: JSON parse fix (getNonStreamingEndpoint strips ?alt=sse via URL parsing)
- [x] GitHub Models MCP non-streaming: stream_options stripped when stream=false

## Stats

69 files changed, +2120/-1867


MayCXC:
unfortunately still giant, but it cuts the diff in half. the heart is still just packages/cli/src/providers/provider-registry.ts and packages/cli/src/providers/provider-definitions.ts and the rest is in response to it.

erudenko:
Closing this, and I want to be clear it isn't because the architecture was wrong. It's because we shipped it.

PR #82 landed on 20 March, one day after you opened this, and its commit message says "Based on architecture from PR #70 by @MayCXC". So main has `createHandlerForProvider` at `provider-profiles.ts:469`, the `PROVIDER_PROFILES` dispatch table, one `resolveModelProvider` plus `createHandlerForProvider` pair in `proxy-server.ts`, and a single-source `BUILTIN_PROVIDERS`. That's the headline of this PR, built from your design.

What it means for the diff is that it can't come back. 69 files, and the four biggest have had 18 to 29 commits each since. Three transports were added that this doesn't know about, one was deleted, and composed-handler grew a whole Layer 4 supervisor.

Two things in it worth flagging anyway.

`ApiKeyTransport.getHeaders()` hardcodes `Authorization: Bearer ${this.apiKey}` in the shared base. That looks like a safe default and it really isn't. Anthropic uses `x-api-key`, Gemini uses `x-goog-api-key`, Devin uses `Basic <k>-<k>` on a lowercase header. A Bearer default doesn't error for any of them, it just authenticates as nobody, and for one provider the documented result is a silent fall-through to OpenRouter. We make each profile state its auth shape now, so a missing one is a compile error instead of a wrong bill.

And the Zen definition swap adds `opencode-zen-minimax` with `transport: "anthropic"` and `apiPath: "/v1/messages"`. We measured that exact route the day before yesterday. Anthropic body to `/v1/messages` gives 401, and with tools it gives 400 invalid tool type. The most recent commit on main is the fix removing that path.

What I do still want from here is `BaseTransport` plus one shared `RequestQueue` with overridable `onResponse`, `shouldRetry` and `calculateDelay`. We're still carrying three queue files, gemini-queue, local-queue and openrouter-queue, doing nearly the same job. `getNonStreamingEndpoint()` with the URL-parsed `?alt=sse` strip is a real fix too. Send either as its own small PR and I'll review it the same week.

erudenko:
Actually closing it now. I wrote the explanation above on 9 August and then never hit the button, so this has been sitting open under a comment saying it was closed. Apologies for the mixed signal.

Everything there stands. `createHandlerForProvider` and the `PROVIDER_PROFILES` dispatch table are on main and I've been working inside them all week, so the architecture from this PR is very much alive even though the diff can't be.

PR #81, opened 2026-03-19
feat: GitHub Models provider, user-defined providers, MCP multi-provider with file context

MayCXC:
## Summary

GitHub Models provider, user-defined providers, MCP multi-provider rewrite with file context. Rebased onto v6.2.2. Depends on #80.

### GitHub Models (`gh@`)
- New builtin provider: `models.github.ai/inference`, OpenAI transport
- Auth: `GITHUB_MODELS_TOKEN` (fine-grained PAT)
- Usage: `gh@gpt-4o`, `gh@o3-mini`

### User-defined providers
- `~/.claudish/config.json` `providers` section
- `registerProvider()` adds to all derived maps (PROVIDERS_BY_NAME, shortcuts, patterns, etc.)
- Called at module init after builtins
- Supports: baseUrl, apiPath, transport, shortcuts, apiKeyEnvVar
- Skips entries that collide with builtin names

### MCP server rewrite
- `runPrompt()` uses the full resolver stack: `resolveEffective` + `resolveTransport` + `resolveAPIFormat` + `adapter.parseResponse()`
- Accepts `provider@model` syntax for any provider: `g@gemini-2.5-flash`, `gh@gpt-4o`, `oai@gpt-5.4`
- Falls back to OpenRouter for bare vendor/model IDs
- Non-streaming: `transport.getNonStreamingEndpoint()` for Gemini (URL-parsed, no ?alt=sse), `stream: false` + `delete stream_options` for OpenAI-compatible
- `files` parameter on `run_prompt` and `compare_models`: file contents appended after the prompt as labeled code blocks
- Removed team-orchestrator tools

### Tested
- [x] `g@gemini-2.5-flash` basic + files
- [x] `gh@gpt-4o` basic + files
- [x] `compare_models` with both providers + files
- [x] 138/138 tests pass


erudenko:
Closing this alongside #80 since it builds on it, but two pieces in here are things I actually want.

The GitHub Models provider is genuinely missing. Nothing matching `models.github.ai` or `GITHUB_MODELS_TOKEN` anywhere in `packages/cli/src`. And your entry is the right shape, `transport: "openai"`, the inference base URL, `/chat/completions`, `shortcuts: ["gh"]`. It drops into `BUILTIN_PROVIDERS` unchanged. That's about 16 lines and I'll take it, after one live curl, since the endpoint and the PAT scope are five months old now.

The `files: string[]` parameter on `run_prompt` and `compare_models` is the other one. Orthogonal to the rest of the PR and it applies cleanly to what we have.

The other two deltas landed differently.

User-defined providers are covered by `customEndpoints` now, with zod validation, `${VAR}` expansion, simple and complex kinds, authScheme, streamFormat, a models allowlist and modelPrefix. One difference worth naming: `runtime-providers.ts:11` says outright that registering must not mutate `BUILTIN_PROVIDERS`, and your `registerProvider()` does `BUILTIN_PROVIDERS.push(def)`. We landed on that rule after a couple of ordering surprises.

MCP multi-provider went another way too. `runPromptViaProxy` at `mcp-server.ts:222` starts the real proxy and POSTs to `/v1/messages`, so it inherits routing, adapters, transports and the behavior layer for free instead of re-resolving all of it by hand.

Two things that would have bitten. `loadUserProviders()` runs at module import with a bare `catch {}`, so a typo in config.json gives you zero providers and no message at all. And `mcp-server.ts` here is -498/+284 with "removed team-orchestrator tools" in the body, while ours is 1407 lines with 11 tools, channel mode and the 1Password spawn pre-hydration. Merging would take channel mode out.

Send the GitHub Models entry as its own PR and it'll be in quickly. Sorry it took this long to give you a straight answer on the stack.

erudenko:
Actually closing now, same as #80. The comment above went up on 9 August and the close never fired, so this sat open under a comment saying otherwise. Sorry.

I re-checked the part I said I wanted: GitHub Models is still missing. Nothing matching `models.github.ai` or `GITHUB_MODELS_TOKEN` anywhere in `packages/cli/src`, and there are 32 builtin providers now with `gh` still free as a shortcut. So that piece is still worth landing on its own.

If you want to open a small PR with just the `BUILTIN_PROVIDERS` entry plus its `PROVIDER_PROFILES` line, I'll take it. Both tables matter — a definition without a profile silently falls through to OpenRouter, which means requests quietly go to a different vendor and get billed there.

Issue #90, opened 2026-03-30
[bug] status line  not displaying correct model and context usage 0


dev-seahouse:
<img width="622" height="481" alt="Image" src="https://github.com/user-attachments/assets/691eb2b0-db1f-4f60-81e3-23d3b04b2154" />

Context:
here is my ~/.claudish/config.json profile config

```
    "default": {
      "name": "default",
      "models": {
        "opus": "zai@zai@glm-5.1",
        "sonnet": "zai@zai@glm-5.1",
        "haiku": "zai@glm-4.7",
        "subagent": "zai@zai@glm-5.1"
      },
      "createdAt": "2026-01-27T15:11:40.474Z",                                                                                                                                                      
      "updatedAt": "2026-03-26T08:44:32.355Z"
    },

```

it is set to glm-5.1, but claudish shows Model: zai@glm5 instead of zai@glm5.1 

also , notice that the context usage in status line is always 100% when using either glm5 or glm5.1

kenan-xin:
the context window for glm are as follows:
```
"models": [
  {
    "id": "glm-5",
    "name": "GLM-5",
    "reasoning": true,
    "input": ["text"],
    "cost": {"input": 0, "output": 0, "cacheRead": 0, "cacheWrite": 0},
    "contextWindow": 204800,
  },
  {
    "id": "glm-4.7",
    "name": "GLM-4.7",
    "reasoning": true,
    "input": ["text"],
    "cost": {"input": 0, "output": 0, "cacheRead": 0, "cacheWrite": 0},
    "contextWindow": 204800,
  },
  {
    "id": "glm-5.1",
    "name": "GLM-5.1",
    "reasoning": true,
    "input": ["text"],
    "cost": {"input": 0, "output": 0, "cacheRead": 0, "cacheWrite": 0},
    "contextWindow": 204800,
  }
]
```
https://docs.z.ai/devpack/using5.1

erudenko:
Two issues identified and fixed:

**1. Wrong context window for GLM models**
The static model catalog had `glm-5` at 80,000 tokens instead of 204,800. This caused the status line to compute incorrect context usage percentages. Fixed to match Z.AI's documented values (204,800 for glm-5, glm-5.1, and glm-4.7).

**2. Missing `glm-5.1` entry**
`glm-5.1` was not in the model catalog at all. It now has its own entry with the correct 204,800 context window.

[...]

PR #92, opened 2026-03-30
fix: add glm-5.1 to model catalog

bluzername:
## Problem

When using `glm-5.1` model via claudish, the status line shows wrong context usage (always 100%) because `glm-5.1` is not in the model catalog. It falls through to the generic `glm-` fallback which has 131k context window - probably not matching the real context window of the model.

The claudish-bot comment on #90 already identified this: "glm-5.1 isn't in the model catalog yet — only glm-5 and glm-4.7 are listed."

## Fix

Added `glm-5.1` entry to the model catalog in `model-catalog.ts`:

```typescript
{ pattern: "glm-5.1", contextWindow: 128_000, supportsVision: true },
```

Placed before `glm-5-turbo` and `glm-5` entries (catalog is ordered by specificity, first match wins) so `glm-5.1` matches exactly instead of falling through to the shorter `glm-5` pattern.

Set `supportsVision: true` same as `glm-5` since the 5.x series supports vision.

Also added a test case in `e2e-model-catalog.test.ts` verifying the new entry returns correct context window and vision support.

## Note

This addresses only part of #90 (the model catalog gap). The other parts (model name display from token file, and usage tracking for Z.AI provider) are separate issues that need investigation with `--debug` logs as the bot suggested.

Ref #90

### DIFF
--- a/packages/cli/src/adapters/model-catalog.ts
+++ b/packages/cli/src/adapters/model-catalog.ts
@@ -39,6 +39,7 @@ export const MODEL_CATALOG: ModelEntry[] = [
   { pattern: "grok", contextWindow: 131_072 },
 
   // ── GLM ─────────────────────────────────────────────
+  { pattern: "glm-5.1", contextWindow: 128_000, supportsVision: true },
   { pattern: "glm-5-turbo", contextWindow: 202_752 },
   { pattern: "glm-5", contextWindow: 80_000, supportsVision: true },
   { pattern: "glm-4.7-flash", contextWindow: 202_752 },
--- a/packages/cli/src/e2e-model-catalog.test.ts
+++ b/packages/cli/src/e2e-model-catalog.test.ts
@@ -50,6 +50,13 @@ describe("Group 1: Model Catalog — lookupModel()", () => {
     expect(entry!.temperatureRange).toBeUndefined();
   });
 
+  test("glm-5.1 → contextWindow 128000, supportsVision true", () => {
+    const entry = lookupModel("glm-5.1");
+    expect(entry).toBeDefined();
+    expect(entry!.contextWindow).toBe(128_000);
+    expect(entry!.supportsVision).toBe(true);
+  });
+
   test("glm-5 → contextWindow 80000, supportsVision true", () => {
     const entry = lookupModel("glm-5");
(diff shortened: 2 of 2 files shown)


kenan-xin:
@bluzername  , the context window for glm5.1 is   204800 

image

source: https://docs.z.ai/devpack/using5.1

builderpepc:
I think you meant to tag @bluzername

erudenko:
Closing this one, and it isn't the code. The file just doesn't exist any more.

`MODEL_CATALOG` is gone from `adapters/model-catalog.ts`. The header there now says the Firebase slim cache is the sole source of truth, and `lookupModel` delegates to `readAllModelsCache()`. The test file you extended was deleted too. So there's no region left to patch and no rebase that would land this.

Where glm-5.1 metadata belongs now is the models-index repo, https://github.com/MadAppGang/models-index. That's the hosted catalog claudish reads at runtime. A data-gap report there is the right move, and it fixes it for everyone rather than only for people running from source.

Worth saying why we moved, since it's the same reason your patch can't apply. Hardcoding a context window means it's wrong the moment a provider changes it, and nobody finds out until someone's context silently truncates. Your `contextWindow: 128_000` would have been right on the day and stale later. Not a criticism of the PR, it was correct for the code that was in front of you in March.

erudenko:
Closing this, but the gap you spotted was real and it's fixed.

`model-catalog.ts` doesn't hold model facts anymore. The header now reads "Firebase slim cache is the sole source of truth" and there are zero hardcoded `pattern:` entries left in the file, so there's nowhere for this entry to go. Model metadata moved out to https://github.com/MadAppGang/models-index and claudish just reads the hosted catalog at runtime.

glm-5.1 already resolves today:

```
$ lookupModel("glm-5.1")
ctx=200000 vision=false
```

Which is also the honest argument for the move. Your patch had 128000 and `supportsVision: true`, and the live catalog says 200000 and false. Not a knock on you, those numbers were probably right somewhere at some point. That's the problem with pinning them in the CLI, they rot quietly and nobody notices until a session compacts 72k early.

If you hit a model that's missing or wrong in the catalog, file it against models-index and it gets fixed for everyone without a claudish release. Thanks for flagging it.

PR #95, opened 2026-04-01
feat: add OpenAI OAuth provider (oo@ prefix)

Caua726:
## Summary

New `openai-oauth` provider (prefix `oo@`) that authenticates via OpenAI's public OAuth PKCE flow, allowing users to use their ChatGPT Plus/Pro subscription without an API key via the Codex backend-api Responses endpoint.

### New files
- `auth/openai-oauth.ts` — PKCE flow, token refresh, secure credential storage
- `providers/transport/openai-oauth.ts` — Bearer auth transport with `store=false`, 429 retry

### Modified files
- `provider-definitions.ts` — `oo@` shortcut, `TransportType` union
- `provider-profiles.ts` — `openaiOAuthProfile` with `CodexAPIFormat`
- `oauth-registry.ts` — credential validation entry
- `index.ts` — `--openai-login` / `--openai-logout` CLI flags
- `cli.ts` — help text + `--models` listing with login status
- `model-selector.ts` — profile provider selector with 7 known Codex models
- `codex-api-format.ts` — `reasoning.effort` mapping from `thinking.budget_tokens`

### Supported models
`gpt-5.4`, `gpt-5.4-mini`, `gpt-5.4-nano`, `gpt-5.3-codex`, `gpt-5.3-codex-spark`, `gpt-5-codex`, `gpt-5-codex-mini`

### Usage
```bash
claudish --openai-login
claudish --model oo@gpt-5.3-codex "task"
claudish --openai-logout
```

### DIFF
--- a/packages/cli/src/adapters/codex-api-format.ts
+++ b/packages/cli/src/adapters/codex-api-format.ts
@@ -13,6 +13,7 @@
 
 import { BaseAPIFormat, type AdapterResult, matchesModelFamily } from "./base-api-format.js";
 import type { StreamFormat } from "../providers/transport/types.js";
+import { log } from "../logger.js";
 
 export class CodexAPIFormat extends BaseAPIFormat {
   constructor(modelId: string) {
@@ -59,6 +60,18 @@ export class CodexAPIFormat extends BaseAPIFormat {
 
     if (claudeRequest.max_tokens) {
       payload.max_output_tokens = Math.max(16, claudeRequest.max_tokens);
+    }
+
+    // Map thinking.budget_tokens -> reasoning_effort for reasoning models
+    if (claudeRequest.thinking) {
+      const { budget_tokens } = claudeRequest.thinking;
+      let effort = "medium";
+      if (budget_tokens < 4000) effort = "low";
+      else if (budget_tokens < 16000) effort = "medium";
+      else if (budget_tokens < 32000) effort = "high";
+      else effort = "high";
+      payload.reasoning = { effort };
+      log(`[CodexAPIFormat] Mapped budget ${budget_tokens} -> reasoning.effort: ${effort}`);
     }
 
     if (tools.length > 0) {
--- a/packages/cli/src/auth/oauth-registry.ts
+++ b/packages/cli/src/auth/oauth-registry.ts
@@ -13,7 +13,7 @@ interface OAuthProviderDescriptor {
  * Providers with working OAuth device authorization flows.
  *
  * Providers NOT listed here use API keys only (no public OAuth device-auth endpoint):
- *   - openai        (OPENAI_API_KEY) - OpenAI does not offer public OAuth device auth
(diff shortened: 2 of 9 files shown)


bassner:
This is already supported through the codex provider, just run ``claudish login codex`` and then use e.g. ``claudish --model cx@gpt-5.5``. Not sure if this is needed

erudenko:
Bad timing on this one and I'm sorry about it. We shipped the same provider on the same day you opened this, `f9b1c54` in v6.5.0 on 1 April. `cx@` with `claudish login codex` does what `oo@` does here. Same PKCE client `app_EMoamEEZ73f0CkXaXp7hrann`, same auth.openai.com endpoints, same scopes, same `chatgpt.com/backend-api/codex/responses`.

So there's nothing to merge. But a couple of things worth saying, because you clearly worked this out yourself rather than copying it from somewhere.

Ours sends `chatgpt-account-id` pulled from the id_token claim, plus `OpenAI-Beta: responses=experimental` and `originator: codex_cli_rs`. Your `OpenAICredentials` interface has no id_token field at all, so I'd expect subscription accounts to have failed auth for you. If you actually got it working against a Plus account I'd like to know how, because that would mean the backend is looser than we think.

The other thing, for future PRs here. `cli.ts` and `model-selector.ts` each pin the same 7 models with literal context windows, and there's a third copy sitting in a comment. We have a rule against that, because per-user rosters drift and hardcoded ones go stale. Yours already has, there's no gpt-5.6 in it. We pull context windows from the live catalog now.

Closing, but thanks for doing the work. Getting a PKCE flow right from scratch isn't nothing.

erudenko:
Closing this, and I'm sorry it's for the least satisfying reason: claudish now ships the same capability under a different prefix, and yours sat here while that happened.

`cx@` / `codex@` is the `openai-codex` provider, and it's the same thing you built:

```ts
name: "openai-codex",
apiPath: "/v1/responses",              // Codex backend Responses endpoint
oauthFallback: "codex-oauth.json",     // ChatGPT Plus/Pro subscription
oauthLoginSlug: "codex",               // claudish login codex
```

The PKCE flow lives in `auth/codex-oauth.ts` and the login is wired through `auth-commands.ts`. So `claudish login codex` then `claudish --model cx@gpt-5.3-codex` is the flow your `oo@` prefix was going to give people.

One difference worth knowing, since it's a design decision rather than an accident: `openai-codex` is deliberately dual-mode. It has `apiKeyAliases: ["OPENAI_API_KEY"]`, so the same provider works on a metered key OR the subscription, whichever the user has. That's also why it's kept out of `SUBSCRIPTION_PROVIDERS` — reporting a flat rate to someone OpenAI is actually metering would silently under-report real money, and there's currently no way to tell which credential is in play from the provider name alone.

743 lines of OAuth is not a small thing to write and I'd rather you hadn't spent it on something that got built in parallel. If `cx@` doesn't cover a case yours did, tell me which and I'll treat that as a gap to fix rather than a PR to close.

PR #101, opened 2026-04-07
fix: strip x-anthropic-billing-header to fix vLLM prefix caching

mtparet:
## Summary

- Strip `x-anthropic-billing-header` from system prompts in `filterIdentity` to fix vLLM prefix caching
- Add unit tests for the identity filter

## Problem

Claude Code injects `x-anthropic-billing-header: cc_version=...; cch=<hash>` into the system prompt on every turn. The `cch` (conversation context hash) changes per turn, making the system prompt tokens differ between requests. This breaks vLLM's prefix cache for the **entire** ~31k token prompt — reducing cache hits from 99% to 0.15% (only 48 tokens cached).

## Root Cause Analysis

1. Sent 4 identical "test" messages via claudish → MiniMax M2.5 on vLLM
2. VictoriaMetrics showed prefix cache metrics were healthy (~50% cluster-wide)
3. But per-request `cached_tokens` was stuck at 48 out of ~31k
4. Dumped full request payloads and diffed between turns
5. Found the only difference: `cch=8ae40` → `cch=4e20a` → `cch=ad80e` in the system prompt

## Fix

Added `.replace(/x-anthropic-billing-header:[^\n]*\n?/g, "")` as the first filter in `filterIdentity()`. This is safe because `filterIdentity` only runs for non-Anthropic providers (OpenAI-compat, Gemini, LiteLLM, OpenRouter, local) — never for native Anthropic passthrough where the header is meaningful.

## Validation

Before fix: `cached_tokens` = 48 / 31k on every turn (0.15%)
After fix: `cached_tokens` = 31,728 / 31,749 by 4th turn (99.9%)

| Turn | Before Fix | After Fix |
|------|-----------|-----------|
| 1st | 48 (0.15%) | 32 (cold) |
| 2nd | 48 (0.15%) | 24,864 (79%) |
| 3rd | 48 (0.15%) | 31,296 (99%) |
| 4th | 48 (0.15%) | 31,728 (99.9%) |

## Test plan

- [x] `bun test identity-filter.test.ts` — 9 tests pass
- [x] `bun run build` — compiles clean
- [ ] Manual: run claudish with `--debug`, send "test" twice, verify `cached_tokens` grows

### DIFF
--- a/packages/cli/src/handlers/shared/format/identity-filter.test.ts
+++ b/packages/cli/src/handlers/shared/format/identity-filter.test.ts
@@ -1,0 +1,64 @@
+import { describe, it, expect } from "bun:test";
+import { filterIdentity } from "./identity-filter.js";
+
+describe("filterIdentity", () => {
+  describe("x-anthropic-billing-header stripping", () => {
+    it("strips billing header with cch hash", () => {
+      const input = "Some system prompt\nx-anthropic-billing-header: cc_version=2.1.92.a35; cc_entrypoint=cli; cch=8ae40;\nMore content";
+      const result = filterIdentity(input);
+      expect(result).not.toContain("x-anthropic-billing-header");
+      expect(result).toContain("More content");
+    });
+
+    it("strips billing header with different cch values", () => {
+      const result1 = filterIdentity("prefix\nx-anthropic-billing-header: cc_version=2.1.92; cch=8ae40;\nsuffix");
+      const result2 = filterIdentity("prefix\nx-anthropic-billing-header: cc_version=2.1.92; cch=4e20a;\nsuffix");
+      // After stripping, both should produce identical output
+      expect(result1).toBe(result2);
+    });
+
+    it("handles header at start of content", () => {
+      const input = "x-anthropic-billing-header: cc_version=2.1.92; cch=abc;\nRest of prompt";
+      const result = filterIdentity(input);
+      expect(result).not.toContain("x-anthropic-billing-header");
+      expect(result).toContain("Rest of prompt");
+    });
+
+    it("handles header at end of content without trailing newline", () => {
+      const input = "Some prompt\nx-anthropic-billing-header: cc_version=2.1.92; cch=abc;";
+      const result = filterIdentity(input);
+      expect(result).not.toContain("x-anthropic-billing-header");
+    });
+
+    it("preserves surrounding content intact", () => {
+      const input = "Line 1\nx-anthropic-billing-header: cc_version=2.1.92; cch=abc;\nLine 3";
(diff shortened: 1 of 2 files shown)


mtparet:
cc @erudenko is the CI failure expected ?

erudenko:
Your measurement is the strongest piece of evidence attached to any PR in this repo right now. 48 tokens cached before, 31,728 after, same prompt. That's the whole argument in one line.

Confirmed the bug is still open on main, `grep -rn "billing-header" packages/cli/src` returns nothing. And your regex is the right one. #126 does the same fix with `cc_version=[^\n]*`, which quietly assumes that token comes first in the header. Yours doesn't assume anything.

What stops me merging it as-is is coverage. `filterIdentity` only gets called from `openai-messages.ts:23` and `gemini-api-format.ts:202`. It never runs for AnthropicApiFormat, which ignores `_filterFn` entirely at `anthropic-api-format.ts:39`, and that's MiniMax, Kimi direct, GLM, Z.AI and vertex-anthropic. It never runs for Codex either, where `payload.instructions = claudeRequest.system` goes through raw, or for OllamaCloud.

Which means your own motivating model is only half covered. `minimax-m2.5` through a custom OpenAI-format vLLM endpoint gets the fix. The same model through `mm@` doesn't.

That's the worst shape for a caching bug. Tests green, looks fixed, still leaking on six providers, and nobody notices because the symptom is a slow bill rather than an error.

So I'm moving it one layer in, to `transform.ts`'s `transformOpenAIToClaude`. That's called only from ComposedHandler and never from NativeHandler, so it covers everything without needing a guard. Your regex and your test file go in with it, and you and jsboige both get credit on the commit.

One more thing worth knowing. The failing check here isn't your code. This is a fork PR and the claude-code workflow wants a repo secret that GitHub won't hand to fork runs. While checking that I noticed the real test gate never ran on this PR at all, only claude-code and GitGuardian, so I'm fixing that separately.

Thanks for doing the measurement. Most people would have just described it.

erudenko:
Shipped this in 228db3c, so I'm closing yours.

Your diagnosis was the part that mattered. The `cch=` token changes every turn and it sits at the head of the system prompt, so it invalidates the backend KV cache on literally every request. I hadn't connected those two facts until I read this.

I put the strip in `transform.ts` rather than `identity-filter.ts`. `transformOpenAIToClaude` is step 1 of the request path for every composed handler, so it covers the custom-endpoint route your vLLM setup takes and both the string and array forms of `system`. One small difference: the array path runs the strip before the existing empty-block filter, so a block that was nothing but the header gets removed instead of left behind as an empty string.

Thanks for finding it.

PR #103, opened 2026-04-09
fix: default to interactive mode when only flags are passed (no prompt)

mtparet:
When claudish receives only flags like --session-id and --dangerously-skip-permissions without a positional prompt argument, it incorrectly defaulted to single-shot mode, adding -p (print mode) to the claude invocation. This caused "Input must be provided either through stdin or as a prompt argument when using --print" errors.

Track whether a positional prompt was found during arg parsing (_hasPositionalPrompt) instead of checking claudeArgs.length, since flag values (e.g. a UUID) don't start with '-' but aren't prompts either.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -366,12 +366,16 @@ export async function parseArgs(args: string[]): Promise<ClaudishConfig> {
       config.claudeArgs.push(arg);
       if (i + 1 < args.length && !args[i + 1].startsWith("-")) {
         config.claudeArgs.push(args[++i]);
+        // The consumed value is a flag argument, not a positional prompt.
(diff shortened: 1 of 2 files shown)


erudenko:
Thanks for this fix @mtparet! The core approach of tracking `_hasPositionalPrompt` instead of checking `claudeArgs.length` is exactly right.

Merged as-is and pushed a follow-up fixup (8932edf) that addresses two edge cases:

1. **`--` separator regression** — args after `--` weren't setting `_hasPositionalPrompt`, so `claudish -- "hello"` would have incorrectly entered interactive mode
2. **Removed `_hasFlagValue`** — it was set but never read, so cleaned it up

Also added 5 test cases covering the full interactive-mode detection matrix. 🎉

PR #104, opened 2026-04-12
fix(#84): support xhigh reasoning_effort for GPT-5.x models

AI-Context-Engineer:
Fixes #84

The previous mapping in `packages/cli/src/adapters/openai-api-format.ts` capped `reasoning_effort` at `high` even when `budget_tokens >= 32000`. Per the openai-python SDK ([`shared/reasoning_effort.py`](https://github.com/openai/openai-python/blob/main/src/openai/types/shared/reasoning_effort.py)), valid values include `xhigh`, which is supported by the newer GPT-5.x reasoning models documented in the issue:

- gpt-5.4, gpt-5.4-pro, gpt-5.4-mini
- gpt-5.3-codex
- gpt-5.2, gpt-5.2-pro, gpt-5.2-codex
- gpt-5.1-codex-max

## Change

New mapping:

```
budget_tokens < 4000    -> minimal
budget_tokens < 16000   -> low
budget_tokens < 32000   -> high
budget_tokens >= 32000  -> xhigh
```

Both `prepareRequest()` and the secondary mapping in the Codex Responses payload builder are updated for consistency.

## Test

When the user runs `claudish --model cx@gpt-5.4 --effort max`:
- Before: Claude Code passes `budget_tokens >= 32000` -> mapped to `high` (loses one tier)
- After: Mapped to `xhigh` (uses the model's full reasoning capacity)

## Notes

This is a minimal fix focused only on extending the upper bound. A more comprehensive fix would include per-model effort tables (issue #84 has the matrix), but that's out of scope for this PR.

### DIFF
--- a/packages/cli/src/adapters/openai-api-format.ts
+++ b/packages/cli/src/adapters/openai-api-format.ts
@@ -37,13 +37,17 @@ export class OpenAIAPIFormat extends BaseAPIFormat {
    * Handle request preparation — reasoning parameters and tool name truncation
    */
   override prepareRequest(request: any, originalRequest: any): any {
-    // Map thinking.budget_tokens -> reasoning_effort for o1/o3 models
+    // Map thinking.budget_tokens -> reasoning_effort for o1/o3/gpt-5.x reasoning models
+    // Valid levels per openai-python SDK (shared/reasoning_effort.py):
+    // none, minimal, low, medium, high, xhigh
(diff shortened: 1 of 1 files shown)


AI-Context-Engineer:
Cerrando: duplica el trabajo de #84 que ya esta abierto y es mas completo. Disculpas por el ruido.

PR #113, opened 2026-04-27
fix: improve prompt cache compatibility for custom Codex endpoints

jujinqian162:
## Summary

  When using a third-party Codex-compatible endpoint via an overridden OpenAI Codex base URL, prompt cache
  hit rates were much lower than when talking to Codex directly.

  This change improves cache compatibility for that setup by:
  - attaching a stable `prompt_cache_key` to Codex Responses API requests
  - sending Codex-compatible request metadata headers for custom Codex transports
  - stripping volatile cache-busting fields from routed Anthropic request payloads before forwarding them
  upstream

  ## Background

  During investigation, one source of instability was request metadata that could vary across otherwise
  equivalent requests.

  In particular, Claude Code injects an `x-anthropic-billing-header` string that may include:
  - a volatile `cch=...` field
  - a more specific `cc_version` value than the upstream service appears to need for cache reuse

  Those values are useful for billing and client attribution, but they can also make semantically identical
  requests look different at the payload level. For third-party Codex-compatible backends, that reduces the
  chance of matching the cache behavior seen when using Codex directly.

  ## What changed

  To make request fingerprints more stable:
  - volatile `cch=...` entries are removed from routed Anthropic payloads
  - `cc_version` values are normalized to a shorter stable form instead of preserving extra patch/build
  suffixes
  - Codex Responses requests now include a stable `prompt_cache_key`
  - Codex-compatible metadata headers are sent on custom Codex transport requests

  Together, these changes make repeated requests look closer to the direct Codex path and reduce avoidable
  cache misses caused by transport-level request variation.

  ## Testing

  - added transport tests for stable `prompt_cache_key` and Codex-compatible headers
  - added sanitizer coverage for volatile cache-buster removal and billing header normalization
  - ran:
    - `bun test packages/cli/src/providers/transport/openai.test.ts packages/cli/src/request-sanitizer.test.ts`

### DIFF
--- a/packages/cli/src/handlers/composed-handler.ts
+++ b/packages/cli/src/handlers/composed-handler.ts
@@ -501,7 +501,7 @@ export class ComposedHandler implements ModelHandler {
 
     // 5c. Provider payload transformation (e.g., CodeAssist envelope wrapping)
     if (this.provider.transformPayload) {
-      requestPayload = this.provider.transformPayload(requestPayload);
+      requestPayload = this.provider.transformPayload(requestPayload, claudeRequest);
     }
 
     // (Middleware beforeRequest ran at step 3b, before buildPayload — see the
--- a/packages/cli/src/providers/transport/openai-codex-cache-key.test.ts
+++ b/packages/cli/src/providers/transport/openai-codex-cache-key.test.ts
@@ -1,0 +1,179 @@
+/**
+ * Regression coverage for OpenAI Codex prompt-cache affinity.
+ *
+ * The cache-routing key must stay stable for a Claude Code conversation across
+ * transport instances and process restarts without exposing the raw session id.
+ * Every request also needs a fallback key, while explicit caller keys and the
+ * existing model/auth payload transforms must continue to win where applicable.
+ *
+ * Hermetic: mock only credentials.getRequestAuth, matching the sibling OAuth
+ * transport test, so no real credentials, filesystem, or network are touched.
+ */
+
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import type { RemoteProvider } from "../../handlers/shared/remote-provider-types.js";
+
+const FAKE_TOKEN = "codex-oauth-token-abc";
+const FAKE_ACCOUNT = "acct-123";
+const CODEX_ENDPOINT = "https://chatgpt.com/backend-api/codex/responses";
+
+const CODEX_OAUTH_AUTH = {
+  headers: {
+    Authorization: `Bearer ${FAKE_TOKEN}`,
+    "OpenAI-Beta": "responses=experimental",
+    originator: "codex_cli_rs",
+    accept: "text/event-stream",
+    "chatgpt-account-id": FAKE_ACCOUNT,
+    "x-conversation-id": "claudish-session",
+    "x-session-id": "claudish-session",
+  },
+  endpoint: CODEX_ENDPOINT,
+  transformPayload: (payload: any) => ({
+    ...payload,
+    store: false,
+    include: ["reasoning.encrypted_content"],
+  }),
+};
+
+let getRequestAuthMock = mock(async (_name: string, _ctx: any) => CODEX_OAUTH_AUTH as any);
+
+mock.module("../../auth/credentials/authority.js", () => ({
+  credentials: {
+    getRequestAuth: (name: string, ctx: any) => getRequestAuthMock(name, ctx),
+  },
+}));
+
+const { OpenAICodexTransport } = await import("./openai-codex.js");
+
+const provider: RemoteProvider = {
+  name: "openai-codex",
+  baseUrl: "https://api.openai.com",
+  apiPath: "/v1/responses",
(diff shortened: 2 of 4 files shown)


erudenko:
The `prompt_cache_key` idea is right and we don't have it. Nothing matching it anywhere in `packages/cli/src`, and `OPENAI_CODEX_BASE_URL` is a supported override, so the third-party Codex backend case you're describing is real.

The diff does three things though, and the title only claims one. So let me go through them.

First, `buildCodexCacheCompatHeaders` gets merged into `getHeaders()` for all Codex requests, including first-party ChatGPT OAuth. It hardcodes `User-Agent: codex_exec/0.120.0 (Ubuntu 22.4.0; x86_64) gnome-terminal` and `originator: codex_exec`. Because the spread puts cacheHeaders last, it overwrites our `originator: codex_cli_rs`, which has a test pinning it. Telling ChatGPT we're a different client on a different OS is a separate decision from prompt caching, and it isn't what the title says.

Second, the sanitizer runs at `/v1/messages` and `/v1/messages/count_tokens` before routing, so it rewrites the system prompt for every provider, native Anthropic passthrough included. `request-sanitizer.ts:47` normalises `cc_version=2.1.7` to `cc_version=2.1` inside Anthropic's own billing header. So we'd be mutating the attribution data we send to Anthropic in order to fix a third-party Codex cache issue. Wrong blast radius. That wants to be a transport-level transform on the Codex path.

Third, the cache key. `buildStablePromptCacheKey` hashes baseUrl plus model plus `process.cwd()`. Two claudish sessions in the same directory on the same model collide on one key and one session_id, and the key outlives unrelated conversations. Per-conversation is the right granularity.

Small one while I'm here: `hadTrailingSemicolon` checks `trimEnd().endsWith("; ")`, which can never be true, since `trimEnd()` has already removed the trailing space. Dead branch.

What I'd take: `prompt_cache_key`, scoped to the Codex transport, keyed per conversation. Drop the client spoofing and the global sanitizer, or re-scope the sanitizer to the custom-endpoint Codex path only. Cut it down to that and I'll review again quickly.

Also `request-sanitizer.test.ts` is committed as one long line with no trailing newline, which will fail the format check.

erudenko:
Checking back, and with an apology attached.

Sixteen fork CI runs were queued unapproved in this repo, some since July, so no test had ever run against your PR. I've cleared the backlog and fork PRs are covered by the gate now. That's a good chunk of why this sat.

The core idea is still right and still missing. Nothing matching `prompt_cache_key` exists anywhere in `packages/cli/src`, and `OPENAI_CODEX_BASE_URL` is a supported override, so the third-party Codex backend case is real.

What I'd take, unchanged from my last comment:

- `prompt_cache_key`, scoped to the Codex transport
- keyed per conversation rather than per `baseUrl + model + cwd`, so two sessions in the same directory don't collide on one key

What I'd want dropped:

- the hardcoded `User-Agent: codex_exec/0.120.0 (Ubuntu 22.4.0; x86_64)` and `originator: codex_exec`, which currently overwrite our `originator: codex_cli_rs` for first-party ChatGPT OAuth too. Telling ChatGPT we're a different client on a different OS is a separate decision from prompt caching and it isn't what the title claims.
- the global sanitizer, or a re-scope of it to the custom-endpoint Codex path. As written it runs before routing, so it rewrites Anthropic's own billing header on native passthrough to fix a third-party cache issue.

Also still there: `hadTrailingSemicolon` tests `trimEnd().endsWith("; ")`, which can never be true since `trimEnd()` already removed the trailing space. And `request-sanitizer.test.ts` is committed as one long line with no trailing newline, which the format check will reject.

Leaving this open. Cut it down to the cache key and I'll review it the same week, which I can now actually promise since the CI gate works.

erudenko:
Rebased, cut down to the cache key, merged. Your commit is on main with your authorship.

Two things changed since my last comment.

The sanitizer half is gone, because it's already solved. `228db3c` (from @mtparet's #101) strips the whole `x-anthropic-billing-header` line in `transformOpenAIToClaude`, so every non-native path including Codex gets a stable prefix, and native Anthropic keeps its own header since that function never runs for `NativeHandler`. Stronger than normalising `cc_version`, and measured: 48 tokens cached before, 31,728 after, on vLLM. So your `cch=` diagnosis was right. It just landed through someone else's PR while this one sat in the queue, which is on me.

And the key is per conversation now. `extractSessionId` in `behavior/harness.ts` already pulls Claude Code's session id out of the `metadata.user_id` JSON blob, and that id is stable across every turn and survives a claudish restart mid-session, neither of which `baseUrl + model + cwd` could do. It's hashed before it goes on the wire. That needed `transformPayload` to see the original Claude body, because the Responses adapter lifts `system` into `instructions` and drops `metadata`, so it takes an optional second arg now. The other two implementers ignore it.

Dropped the `codex_exec` User-Agent and originator, for the reason in my earlier comment.

Also worth saying plainly: the idea was right and nothing matching `prompt_cache_key` existed anywhere in `packages/cli/src`. You found a real gap on a path most people never touch, and then waited four months on a CI queue that was never going to run. Thanks for the patience, and sorry it took this long.

PR #123, opened 2026-05-16
fix(stream-parsers): forward real input_tokens in message_delta for autocompact

### LINKED ISSUE #90: [bug] status line  not displaying correct model and context usage 0

jsboige:
## Summary

- Fix hardcoded `input_tokens: 100` in `message_start` and missing `input_tokens` in `message_delta` across all 4 non-Anthropic stream parsers
- This breaks Claude Code's autocompact/condensation for every non-Anthropic provider (GLM, Qwen, Gemini, Ollama, Codex, OpenRouter, LiteLLM)

## Problem

Claude Code accumulates token usage from SSE events via `updateUsage()` in `claude.ts`:
1. `message_start` seeds `usage.input_tokens = 100` (hardcoded placeholder)
2. `message_delta` sends `{ output_tokens: N }` but **never includes `input_tokens`**
3. `updateUsage()` only replaces `input_tokens` when the new value is `> 0` — since it's never sent, it stays at 100 forever
4. `tokenCountWithEstimation()` computes ~30k instead of ~80k for a typical session
5. Autocompact threshold is never reached → context grows unbounded → agents stall/crash

## Root Cause

```
message_start:  usage: { input_tokens: 100, output_tokens: 1 }   ← hardcoded
message_delta:  usage: { output_tokens: <real> }                  ← missing input_tokens
```

The real `input_tokens` (from `state.usage.prompt_tokens` / `usageMetadata.promptTokenCount` / `prompt_eval_count`) is available in the parser but never forwarded to Claude Code.

## Changes

| File | Before | After |
|------|--------|-------|
| openai-sse.ts:141 | `input_tokens: 100` | `input_tokens: 0` |
| openai-sse.ts:296 | `{ output_tokens }` | `{ input_tokens, output_tokens }` |
| openai-sse.ts:315 | `onTokenUpdate(100, ...)` | `onTokenUpdate(0, ...)` |
| gemini-sse.ts:65 | `input_tokens: 100` | `input_tokens: 0` |
| gemini-sse.ts:116 | `{ output_tokens }` | `{ input_tokens, output_tokens }` |
| ollama-jsonl.ts:53 | `input_tokens: 100` | `input_tokens: 0` |
| ollama-jsonl.ts:78 | `{ output_tokens }` | `{ input_tokens, output_tokens }` |
| openai-responses-sse.ts:69 | `input_tokens: 100` | `input_tokens: 0` |

Note: `openai-responses-sse.ts` already sent `input_tokens` in its `message_delta` — only the `message_start` placeholder needed fixing.

## Validation

- `bun run build` — compiles clean
- Deployed on our cluster for 24h — autocompact triggers correctly at the configured threshold (80%)
- `/context` now shows accurate token counts instead of "100 / 200k (0%)"

Fixes #90

## Test plan

- [ ] `bun run build` passes
- [ ] Start proxy with a non-Anthropic model, send messages, verify `/context` shows accurate usage
- [ ] Verify autocompact triggers at the expected threshold
- [ ] Run `bun test packages/cli/src/format-translation.test.ts` — SSE replay tests pass

### DIFF
--- a/packages/cli/src/handlers/shared/stream-parsers/gemini-sse.ts
+++ b/packages/cli/src/handlers/shared/stream-parsers/gemini-sse.ts
@@ -62,7 +62,7 @@ export function createGeminiSseStream(
           model: opts.modelName,
           stop_reason: null,
           stop_sequence: null,
-          usage: { input_tokens: 100, output_tokens: 1 },
+          usage: { input_tokens: 0, output_tokens: 0 },
         },
       });
       send("ping", { type: "ping" });
@@ -113,7 +113,7 @@ export function createGeminiSseStream(
           send("message_delta", {
             type: "message_delta",
             delta: { stop_reason: hasToolCalls ? "tool_use" : "end_turn", stop_sequence: null },
-            usage: { output_tokens: outputTokens },
+            usage: { input_tokens: inputTokens, output_tokens: outputTokens },
           });
           send("message_stop", { type: "message_stop" });
         }
--- a/packages/cli/src/handlers/shared/stream-parsers/ollama-jsonl.ts
+++ b/packages/cli/src/handlers/shared/stream-parsers/ollama-jsonl.ts
@@ -50,7 +50,7 @@ export function createOllamaJsonlStream(
           model: opts.modelName,
           stop_reason: null,
           stop_sequence: null,
-          usage: { input_tokens: 100, output_tokens: 1 },
+          usage: { input_tokens: 0, output_tokens: 0 },
         },
       });
       send("ping", { type: "ping" });
@@ -75,7 +75,7 @@ export function createOllamaJsonlStream(
           send("message_delta", {
             type: "message_delta",
             delta: { stop_reason: "end_turn", stop_sequence: null },
-            usage: { output_tokens: completionTokens },
+            usage: { input_tokens: promptTokens, output_tokens: completionTokens },
           });
           send("message_stop", { type: "message_stop" });
         }
--- a/packages/cli/src/handlers/shared/stream-parsers/openai-responses-sse.ts
+++ b/packages/cli/src/handlers/shared/stream-parsers/openai-responses-sse.ts
@@ -66,7 +66,7 @@ export function createResponsesStreamHandler(
           model: opts.modelName,
           stop_reason: null,
           stop_sequence: null,
-          usage: { input_tokens: 100, output_tokens: 1 },
+          usage: { input_tokens: 0, output_tokens: 0 },
         },
       });
       send("ping", { type: "ping" });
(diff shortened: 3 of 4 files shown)


erudenko:
Closing this one, but the diagnosis in it was right and it's already fixed on main.

We landed it in `ae8c07f` (v7.18.2) as a shared `messageStartUsage(priorInputTokens)` helper in `stream-parsers/message-start-usage.ts`, applied across every parser instead of per-parser. Your finding that `updateUsage()` only replaces `input_tokens` when the delta is greater than zero is the exact mechanism that fix is built on. That was the hard part, and you found it.

There's a second reason not to merge this now, beyond the conflict. Your version seeds `message_start` with `input_tokens: 0` and then sends `input_tokens: 0` unconditionally in `message_delta`. Main carries the previous turn's count forward instead, with 100 only as a first-turn fallback, and omits the key entirely when the value is zero. Because of the greater-than-zero rule you documented, a zero seed means any turn where the backend skips usage reports an empty context. Same disarmed compaction, one layer down. Both halves are pinned now at `usage-regressions.test.ts:166` and `:218`.

We've since added a fifth parser, `devin-connect.ts`, with the same contract. The shared-helper shape is holding, which I think says your read of the problem was the correct one.

Closing #90 along with this. The write-up saved us real time.

erudenko:
This is fixed on main, so I'm closing it. Your diagnosis is sitting in the source as a comment:

```ts
// input_tokens must ride the delta too: Claude Code takes the
// context size from the last assistant message, and message_start
// could only carry an estimate. Omitting it left the client
// believing every conversation was 100 tokens, which silently
// disabled auto-compaction on every openai-sse provider.
```

That's `openai-sse.ts:420`. Same conclusion, right down to the 100.

Two differences worth knowing. The hardcoded `input_tokens: 100` in `message_start` is gone, but it went to `priorInputTokens` (the previous request's real context size from TokenTracker) rather than to 0 like your patch. A parser genuinely can't know the count until the stream ends, and Claude Code keeps whatever `message_start` carried whenever the closing delta omits it, so a real prior number degrades better than a zero.

And `input_tokens` now only rides the delta when the provider actually reported `prompt_tokens`, instead of always. Sending a hard 0 would overwrite that seed with something worse than the estimate.

All four parsers carry it now, not just the ones in your diff.

This one was a genuinely good find. Auto-compaction silently not firing is the kind of bug people live with for months and blame on the model. Thanks.

PR #124, opened 2026-05-16
fix(routing): recognize runtime provider credentials in routing

jsboige:
## Summary

- Custom endpoints registered at runtime via `registerRuntimeProvider()` have credentials validated at load time
- But `hasCredentialsForProvider()` didn't know about them, so routing excluded them from the candidate chain
- Fix: check `getRuntimeProviders()` before falling through to env var checks
- Without this fix, custom endpoints with valid API keys in config are excluded from routing

## Changes

One file touched: `packages/cli/src/providers/routing-rules.ts`

- Added import of `getRuntimeProviders` from `runtime-providers.js`
- Added early-return check in `hasCredentialsForProvider()` that recognizes runtime-registered providers as always credentialed (the loader validates `apiKey` presence before registration)

## Test plan

- [ ] Configure a custom endpoint in `~/.claudish/config.json` under `customEndpoints` with a valid API key
- [ ] Run `claudish --probe <custom-endpoint-model>` — should show the custom endpoint in the routing chain instead of `no-route`
- [ ] Run `claudish --model <custom-endpoint>@<model> "hello"` — should route successfully without credential errors

### DIFF
--- a/packages/cli/src/providers/routing-rules.ts
+++ b/packages/cli/src/providers/routing-rules.ts
@@ -12,6 +12,7 @@ import {
   isProviderAvailable,
 } from "./provider-definitions.js";
 import { buildCredentialHint } from "./routing-hints.js";
+import { getRuntimeProviders } from "./runtime-providers.js";
 
 /**
  * Pure merge — defaults < global < local. Exposed for testability so callers
@@ -216,6 +217,10 @@ export type RoutePlan =
  */
 function hasCredentialsForProvider(provider: string): boolean {
   if (isLocalTransport(provider)) return isLocalProviderEnabled(provider);
+
+  // Custom endpoints registered at runtime always have credentials —
+  // the loader validates apiKey presence before registration.
+  if (getRuntimeProviders().has(provider)) return true;
 
   if (provider === "native-anthropic") {
     return !!process.env.ANTHROPIC_API_KEY || !!process.env.ANTHROPIC_AUTH_TOKEN;


erudenko:
Closing this, fixed on main but by a different route.

`hasCredentialsForProvider` is now a one-liner that delegates to the credential authority, `credentials.isAvailable(provider)` at `routing-rules.ts:243`. The inline env-var body you patched is gone. Custom endpoints register into that same authority at `custom-endpoints-loader.ts:75` with a `declaredKey` callback, so a config-declared key counts as a real credential and the routing chain stops filtering those endpoints out.

The bug was real and you read it correctly. The one thing I'd have pushed back on is `getRuntimeProviders().has(provider) → return true`, which effectively says registered therefore credentialed. The loader validates the shape of the config, not that `${VAR}` actually expanded. So an endpoint with an unset env var would sail through the pre-flight and then fail at request time, which is a worse failure than the one you were fixing. Main resolves the key for real instead.

Thanks for finding it.

erudenko:
Fixed on main, so closing. Verified through the real loader path:

```
loadCustomEndpoints({ "my-vllm": {apiKey: "${MY_VLLM_KEY}"}, "lit-ep": {apiKey: "sk-literal-abc"} })
my-vllm credentialed = true
lit-ep  credentialed = true
```

It landed somewhere different from your patch though, and I think the difference matters.

Yours makes `hasCredentialsForProvider()` treat any runtime-registered provider as credentialed by definition. That fixes the routing exclusion, but it also means a custom endpoint whose `${VAR}` is unset still counts as available, so instead of being skipped in the chain it gets picked and fails later at the request. Trading a silent exclusion for a late failure.

What's on main instead: `custom-endpoints-loader.ts` calls `credentials.registerApiKeyProvider()` when it registers the endpoint, so the endpoint lives in the same credential authority as every builtin and gets asked the same question. Routing didn't need to learn anything about runtime providers, it just stopped having a blind spot.

The comment on that method describes your exact symptom, including that literal and `${VAR}` keys were the cases that broke:

> every `${VAR}`/literal endpoint failed the routing pre-flight with "No API key for provider"

Worth admitting: I first reproduced this as still-broken and was about to tell you it was live. My repro called `registerRuntimeProvider()` directly, which skips the loader, and the loader is where the credential registration happens. So I'd written a test that omitted the fix. Going through `loadCustomEndpoints` gave the right answer.

Good report. Custom endpoints being invisible to routing is a nasty one because nothing errors, the provider just quietly isn't there.

PR #125, opened 2026-05-16
fix(proxy): transparent header passthrough for Anthropic Max subscription

jsboige:
## Summary

The proxy currently forwards only three specific headers (`authorization`, `x-api-key`, `anthropic-beta`) to `api.anthropic.com`. Anthropic Max subscription auth requires additional internal headers that Claude Code sends — when the proxy drops them, Max subscription users get auth failures on Opus/Sonnet.

This PR changes the proxy to forward **all incoming headers** transparently, filtering out only standard hop-by-hop headers (`host`, `connection`, `transfer-encoding`, `keep-alive`, `te`, `trailer`, `upgrade`, `content-length`) per RFC 2616 section 13.5.1.

### Changes

- **`native-handler.ts`**: Replace selective 3-header forwarding with full passthrough via `Object.entries(originalHeaders)` filtered through a hop-by-hop exclusion set. The advisor beta-stripping logic is preserved but simplified (now only runs when `advisorSwapped` is true, and uses `delete` when stripped is empty instead of leaving the original header).
- **`proxy-server.ts`**: Apply the same transparent passthrough to the `count_tokens` endpoint (was only sending `Content-Type` + `x-api-key`).

### Why not selective forwarding?

Max subscription auth involves headers beyond the three currently forwarded. Rather than playing whack-a-mole adding individual headers as new ones are discovered, forwarding everything (minus hop-by-hop) is the correct proxy behavior and matches what a simple reverse proxy would do.

### No breaking changes

Existing auth flows (API key, OAuth token, direct Anthropic key) all continue to work — they are simply forwarded as-is now instead of being selectively picked.

## Test plan

- [x] Verified diff against `upstream/main` contains only the header forwarding change (no proxyKey, no fork-specific code)
- [x] `native-handler.ts` change: 58 lines changed, purely header construction logic
- [x] `proxy-server.ts` change: 15 lines changed, count_tokens endpoint only
- [ ] Manual test: run `claudish` as proxy with Anthropic Max subscription credentials, confirm Opus/Sonnet requests succeed
- [ ] Manual test: confirm `anthropic-beta` header still passes through for tool use
- [ ] Manual test: confirm count_tokens endpoint still works

### DIFF
--- a/packages/cli/src/handlers/native-handler.ts
+++ b/packages/cli/src/handlers/native-handler.ts
@@ -139,40 +139,42 @@ export class NativeHandler implements ModelHandler {
     log(`Request body (Model: ${target}):`);
     log("=== End Request ===\n");
 
-    // Build headers - pass through auth headers exactly as received
+    // Transparent passthrough: forward ALL incoming headers to api.anthropic.com.
+    // This is essential for Max subscription auth — Claude Code sends internal
+    // headers that make the subscription work for Opus/Sonnet, and we must not
+    // drop any of them.
+    //
+    // Skip hop-by-hop headers (per RFC 2616 §13.5.1) and content-length (the
+    // body is re-serialized below).
+    const HOP_BY_HOP = new Set([
+      "host", "connection", "keep-alive", "transfer-encoding", "te",
+      "trailer", "upgrade", "content-length",
+    ]);
     const headers: Record<string, string> = {
       "Content-Type": "application/json",
-      "anthropic-version": originalHeaders["anthropic-version"] || "2023-06-01",
     };
+    for (const [key, value] of Object.entries(originalHeaders)) {
+      if (HOP_BY_HOP.has(key.toLowerCase())) continue;
+      if (typeof value !== "string") continue;
+      headers[key] = value;
+    }
 
-    // Pass through auth headers as-is
-    if (originalHeaders["authorization"]) {
-      headers["authorization"] = originalHeaders["authorization"];
-    }
-    if (originalHeaders["x-api-key"]) {
-      headers["x-api-key"] = originalHeaders["x-api-key"];
-    }
-    if (originalHeaders["anthropic-beta"]) {
+    // Advisor-swap: strip advisor beta flag when we swapped the tool
+    if (advisorSwapped && originalHeaders["anthropic-beta"]) {
       const incomingBeta = originalHeaders["anthropic-beta"];
-      if (advisorSwapped) {
-        // When we swap the advisor tool we must also strip the matching beta
-        // flag; otherwise Anthropic rejects the request (beta enabled but no
-        // matching server tool declared).
-        const { stripped, changed } = stripAdvisorBeta(incomingBeta);
-        if (changed) {
-          log(
-            `[Native][advisor-swap] stripped advisor-tool beta; before=${incomingBeta} after=${stripped ?? "(empty)"}`
-          );
-          logAdvisorEvent(advisorCfg, {
-            kind: "beta_stripped",
-            before: incomingBeta,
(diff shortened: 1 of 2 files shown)


erudenko:
I want to push back on this one before it goes further, because I can't find the bug it's fixing.

Main already forwards `authorization` at `native-handler.ts:200`, `x-api-key` at `:203` and `anthropic-beta` at `:219`. As far as I know that's the whole Max and OAuth set. The description says unlisted internal headers get dropped and cause auth failures, but it doesn't name one. Can you send a capture of a real Max-subscription request header set, diffed against those? If something genuinely is missing I'll take the change, I just can't verify it from here.

Two things in the diff would regress today regardless of that.

It deletes `"anthropic-version": originalHeaders["anthropic-version"] || "2023-06-01"`. Any client that doesn't send that header now 400s, and `--probe` is one of them. There's a comment at `native-handler.ts:196` saying we accommodate that on purpose.

The count_tokens hunk drops `if (anthropicApiKey) headers["x-api-key"] = anthropicApiKey` at `proxy-server.ts:836`. That's the only auth source when the client sends none.

Worth knowing too, that header block was rewritten since you opened this. It now has a credential-authority fallback at `:207` for the no-inbound-auth case, and this diff would revert it.

The advisor change is good though. Deleting `anthropic-beta` when `stripped` comes back empty, rather than sending an empty header, is right. I'd take that on its own tomorrow.

Blanket header forwarding is arguably correct proxy behaviour and I'm not against it in principle. But it's a behaviour change with no tests, so it wants to be its own PR rather than riding along with a bug fix.

erudenko:
The problem is real, I don't want the fix as written. Max users getting auth failures on a passthrough that's supposed to be transparent is a bad look and I'd like it fixed.

Two blockers.

**Blanket forwarding undoes a deliberate strip.** `native-handler.ts:220` doesn't just pass `anthropic-beta` through, it conditionally rewrites it:

```ts
if (advisorSwapped) {
  const { stripped, changed } = stripAdvisorBeta(incomingBeta);
  ...
  if (stripped) headers["anthropic-beta"] = stripped;
}
```

When we swap the advisor tool we have to strip the matching beta flag, because Anthropic rejects a request that enables a beta with no matching server tool declared. Forwarding every header verbatim puts the original flag back and breaks advisor-swap requests. Silently, and only for people using that path.

**And I don't want to forward everything to api.anthropic.com on principle.** Hop-by-hop filtering per RFC 2616 is the right idea for a proxy sitting between two parties who both know they're proxied. This one has a browser-ish client on one side, so cookies, `x-forwarded-*`, and anything else Claude Code or a future version decides to send all get relayed to a third party. Header forwarding should be a decision per header, not a default.

What I'd merge instead: keep the current explicit list and add the specific headers Max needs, with the beta branch untouched. Something like an `ANTHROPIC_FORWARD_HEADERS` allowlist next to the existing three, so the next missing header is a one-line addition rather than another architecture change.

To do that I need to know which headers those are. Can you get a capture from a Max session? `claudish --debug` logs the inbound set, or a proxy trace works too. Name the headers and I'll wire them in, or take your PR if you narrow it to those.

I don't have a Max subscription on this account, so I can't reproduce the failure or verify a fix myself. That's the real reason I'm asking rather than guessing at the list.

jsboige:
Closing this PR — your review was right and I can't satisfy the 'name the headers' requirement without a Max subscription capture, which I don't have access to. The advisor beta-strip fix from the original commit is a no-op against upstream's current architecture (the header is only set when `stripped` is non-empty), so there's nothing narrow to land here. If a Max user surfaces with a specific header list in the future, the narrow PR becomes addition-only.

PR #126, opened 2026-05-16
fix(proxy): strip Claude Code billing header from prompt for non-Anthropic providers

jsboige:
## Summary

Strips the `x-anthropic-billing-header: cc_version=...; cch=XXXXX;` line that Claude Code injects into the system prompt body when the request is **not** routed to Anthropic's native API.

## Problem

Claude Code injects a billing header into the `system` field of every `/v1/messages` request. The `cch=` token changes on every request, which **breaks prefix caching on self-hosted inference engines** (vLLM, Ollama, LM Studio) that use strict hash matching for KV cache hits. Each request is treated as a unique prefix, resulting in zero cache hits and wasted GPU compute re-processing the same system prompt repeatedly.

## Fix

- In the `/v1/messages` handler, after resolving the handler and before forwarding, check if the handler is a `NativeHandler` (Anthropic direct).
- If **not** native (i.e., OpenRouter, local models, custom endpoints, etc.), strip the billing header line from `body.system` using a regex.
- Handles both `string` and `array` (block) forms of the `system` field per the Anthropic API spec.
- **Zero impact on Anthropic routing** — native requests pass through unchanged.

## Testing

- Verified diff scope: only `packages/cli/src/proxy-server.ts` touched (23 insertions, 0 deletions).
- The `NativeHandler` import already exists in the file — no new dependencies.

### DIFF
--- a/packages/cli/src/proxy-server.ts
+++ b/packages/cli/src/proxy-server.ts
@@ -571,6 +571,29 @@ export async function createProxyServer(
       const body = await c.req.json();
       const handler = await getHandlerForRequest(body.model);
 
+      // Strip Claude Code billing header from system prompt for non-Anthropic
+      // providers. Claude Code injects `x-anthropic-billing-header: cc_version=...; cch=XXXXX;`
+      // into the prompt body — the `cch=` token changes every request, breaking
+      // prefix caching on self-hosted inference (vLLM, Ollama). Only Anthropic
+      // needs this header.
+      if (!(handler instanceof NativeHandler)) {
+        if (typeof body.system === "string") {
+          body.system = body.system.replace(
+            /x-anthropic-billing-header: cc_version=[^\n]*\n?/g,
+            ""
+          );
+        } else if (Array.isArray(body.system)) {
+          for (const block of body.system) {
+            if (block.type === "text" && typeof block.text === "string") {
+              block.text = block.text.replace(
+                /x-anthropic-billing-header: cc_version=[^\n]*\n?/g,
+                ""
+              );
+            }
+          }
+        }
+      }
+
       // Route
       return handler.handle(c, body);
     } catch (e) {


mtparet:
We have the same fix https://github.com/MadAppGang/claudish/pull/101 this is really a must have otherwise claudish is not usable because it cost 10X @erudenko

erudenko:
This is going in, just not quite as written. Leaving it open until it does.

The bug is real and nobody had touched it. `grep -rn "billing-header" packages/cli/src` returns nothing on main. We even have the evidence sitting in the repo already, `evidence-req-advisor-enabled.json` has `system[0]` as the literal `x-anthropic-billing-header: cc_version=2.1.107.3d9; cc_entrypoint=cli; cch=74943;`. That `cch=` changes every turn and sits at the head of a 31k prompt, so every prefix-caching backend re-reads the lot.

#101 fixes the same thing at `filterIdentity`, with 9 tests and a measurement: 48 tokens cached before, 31,728 after. You picked the better layer, they have the better regex and the tests. `cc_version=[^\n]*` assumes that token comes first in the header, `[^\n]*` doesn't assume anything.

I'm going to land it a bit deeper than either of you, at `transform.ts`'s `transformOpenAIToClaude`. That runs for every ComposedHandler and never for NativeHandler, so it covers the anthropic-passthrough providers (MiniMax, Kimi, GLM, Z.AI), plus Codex and OllamaCloud, none of which `filterIdentity` reaches. It also means no `instanceof NativeHandler` guard, the structure gives you that for free. And it's already the place where the system array gets normalised and empty-filtered, so deleting the block instead of blanking it is one line there.

On the blanking, by the way. `{type:"text", text:""}` is harmless today only because `transform.ts:384` drops empty strings when it joins, which you had no way of knowing. On an Anthropic-compat endpoint an empty text block is a 400.

You and mtparet both get credit on the commit. I'll close this and #101 once it's in.

erudenko:
Same fix landed in 228db3c, so I'm closing this one.

You did it at the request boundary in `proxy-server.ts` gated on `!(handler instanceof NativeHandler)`, I did it in `transform.ts`. Two practical differences. The `transform.ts` version doesn't need the handler-type check, since native Anthropic never goes through that path anyway. And the array branch strips before the existing empty-block filter, so a text block containing only the header is dropped rather than turned into an empty string.

Your regex is anchored on `cc_version=` where mine matches the header name alone. Yours is tighter, but Claude Code has already changed the field set once, so I went with the looser one.

You and @mtparet arrived at this independently in #101, which is usually a good sign the bug is real. Thanks for the report.

PR #127, opened 2026-05-16
fix(stream-parsers): clamp content block indices in Anthropic SSE passthrough

jsboige:
## Summary

- Add content block index tracking and clamping in the Anthropic SSE passthrough stream parser
- Prevents "Content block not found" errors when providers (z.ai, MiniMax) send non-sequential content block indices
- Handles both filtered (thinking block suppression) and unfiltered code paths

## Problem

Some Anthropic-compatible providers (z.ai, MiniMax, Kimi) occasionally send SSE events with content block indices that jump or are non-sequential (e.g., 0 → 2, skipping 1). Claude Code expects sequential indices (0, 1, 2, ...) and throws "Content block not found" when it encounters a gap.

Additionally, when `filterThinking` is enabled and thinking blocks are suppressed, the re-indexing logic now properly tracks and clamps indices to prevent the same issue.

## Changes

- **Index tracking**: `highestSeenIndex` tracks the highest content block index seen so far
- **`clampIndex()`**: If an incoming index exceeds `highestSeenIndex + 1`, it's clamped down to the expected sequential value
- **`trackIndex()`**: Updates `highestSeenIndex` on `content_block_start` events
- **Filtered path**: Re-indexed blocks (after thinking suppression) are also clamped
- **Unfiltered path**: Both `content_block_start` remapping and delta/stop clamping applied
- **Logging**: Diagnostic logs when index jumps are detected, useful for debugging provider quirks

## Test plan

- [ ] Verify existing format translation tests still pass: `bun test packages/cli/src/format-translation.test.ts`
- [ ] Manual: `claudish --model zai/claude-sonnet-4-20250514 "use webReader to fetch https://example.com"` — should complete without "Content block not found"
- [ ] Manual: test with a thinking-enabled model that uses `filterThinking` — verify re-indexing is correct
- [ ] Check debug logs for `[AnthropicSSE] Index jump detected` messages when using z.ai or MiniMax

### DIFF
--- a/packages/cli/src/format-translation.test.ts
+++ b/packages/cli/src/format-translation.test.ts
@@ -503,8 +503,12 @@ describe("Regression: Anthropic SSE abandoned streams emit a terminal tail", () 
     });
 
     const events = await parseClaudeSseStream(response);
+    // Under index clamping (see anthropic-sse index-clamping tests), the
+    // client sees the block at the renumbered sequential index (0), not the
+    // upstream index (3) — so the synthetic tail must close the index the
+    // client actually has open.
     const blockStopIndex = events.findIndex(
-      (event) => event.data?.type === "content_block_stop" && event.data?.index === 3
+      (event) => event.data?.type === "content_block_stop" && event.data?.index === 0
     );
     const messageStopIndex = events.findIndex((event) => event.data?.type === "message_stop");
 
--- a/packages/cli/src/handlers/shared/stream-parsers/anthropic-sse-index-clamping.test.ts
+++ b/packages/cli/src/handlers/shared/stream-parsers/anthropic-sse-index-clamping.test.ts
@@ -1,6 +1,14 @@
 import { describe, expect, it } from "bun:test";
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
 
 import { createAnthropicPassthroughStream } from "./anthropic-sse.js";
+
+const FIXTURES_DIR = join(
+  dirname(fileURLToPath(import.meta.url)),
+  "../../../test-fixtures/sse-responses"
+);
 
 const ctx: any = {
   body: (stream: any, init: any) => new Response(stream, init),
@@ -63,6 +71,25 @@ const messageDelta = () =>
   });
 
 const messageStop = () => frame("message_stop", { type: "message_stop" });
+
+/** Replay a whole fixture file (stripping its `# ` metadata lines) through the parser. */
+const runFixture = async (name: string): Promise<string> => {
+  const text = readFileSync(join(FIXTURES_DIR, name), "utf-8")
+    .split("\n")
+    .filter((l) => !l.startsWith("# "))
+    .join("\n");
+  return stripPingFrames(await run([text]));
+};
+
+/** Parse a parser-emitted wire into its data payloads, in order. */
+const parseEmitted = (wire: string): any[] =>
+  wire
+    .split("\n\n")
+    .flatMap((frame) => frame.split("\n"))
+    .filter((l) => l.startsWith("data: "))
+    .map((l) => l.slice(6))
+    .filter((d) => d !== '{"type":"ping"}')
+    .map((d) => JSON.parse(d));
 
 describe("anthropic-sse content block index clamping", () => {
   it("passes sequential indices through untouched", async () => {
@@ -161,5 +188,50 @@ describe("anthropic-sse content block index clamping", () => {
     // Sequential: exactly one block start, one stop, in order.
     expect(out.indexOf("content_block_start")).toBeLessThan(out.indexOf("text_delta"));
     expect(out.lastIndexOf("content_block_stop")).toBeGreaterThan(out.indexOf("text_delta"));
+  });
+});
+
+describe.each([
+  { reqN: "r10324", textFragment: "Search-4-LocalSearch", toolFragment: "check_twin_parity.py" },
+  { reqN: "r10416", textFragment: "QuantConnect", toolFragment: "config.json" },
+])("production fixture: MiniMax-M3 implicit signature block ($reqN)", ({ reqN, textFragment, toolFragment }) => {
+  it("renumbers the whole stream sequentially and drops the implicit signature block", async () => {
+    // Fixtures reconstructed from production captures (see the file header for
+    // the reconstruction proof): MiniMax emits a signature_delta + stop at
+    // index 0 with NO content_block_start (implicit block), then the text
+    // block at index 1 and the tool_use block at index 2.
+    const events = parseEmitted(await runFixture(`minimax-m3-anthropic-implicit-signature-${reqN}.sse`));
+
+    // The implicit signature block never existed client-side: dropped whole.
+    expect(events.filter((e) => e.type === "content_block_delta" && e.delta?.type === "signature_delta")).toHaveLength(0);
+
+    // Exactly two blocks are opened, renumbered sequentially: text at 0, tool_use at 1.
+    const starts = events.filter((e) => e.type === "content_block_start");
+    expect(starts.map((e) => [e.index, e.content_block.type])).toEqual([
+      [0, "text"],
+      [1, "tool_use"],
+    ]);
+
+    // Every text delta follows the remap: all at 0, none leaked at 1.
+    const textDeltas = events.filter((e) => e.delta?.type === "text_delta");
+    expect(textDeltas.length).toBeGreaterThan(5);
+    expect(new Set(textDeltas.map((e) => e.index))).toEqual(new Set([0]));
+    expect(textDeltas.map((e) => e.delta.text).join("")).toContain(textFragment);
+
+    // The tool call's input stream follows its own remap: all at 1, none at 2,
+    // and the concatenated partial_json reassembles into valid JSON.
+    const toolDeltas = events.filter((e) => e.delta?.type === "input_json_delta");
+    expect(toolDeltas.length).toBeGreaterThanOrEqual(1);
+    expect(new Set(toolDeltas.map((e) => e.index))).toEqual(new Set([1]));
+    const toolInput = JSON.parse(toolDeltas.map((e) => e.delta.partial_json).join(""));
+    expect(JSON.stringify(toolInput)).toContain(toolFragment);
+
+    // Blocks close in order, the terminal pair is present, and no index ever
+    // escapes the sequential range.
+    expect(events.filter((e) => e.type === "content_block_stop").map((e) => e.index)).toEqual([0, 1]);
+    expect(events.at(-1)?.type).toBe("message_stop");
+    const stopReason = events.find((e) => e.type === "message_delta")?.delta?.stop_reason;
+    expect(stopReason).toBe("tool_use");
+    expect(events.filter((e) => typeof e.index === "number" && e.index > 1)).toHaveLength(0);
   });
 });
--- a/packages/cli/src/handlers/shared/stream-parsers/anthropic-sse-index-clamping.test.ts
+++ b/packages/cli/src/handlers/shared/stream-parsers/anthropic-sse-index-clamping.test.ts
@@ -104,7 +104,10 @@ describe("anthropic-sse content block index clamping", () => {
     expect(out).toContain("second");
   });
 
-  it("clamps a delta whose index jumps past the last known block", async () => {
+  it("clamps an orphan delta to the last OPEN block, not the next slot", async () => {
+    // A delta can only reference a block the client has opened. Clamping it
+    // to highest + 1 would emit a delta for an unopened block — the same
+    // "Content block not found" this guard exists to prevent.
     const frames = [
       messageStart(),
       textBlockStart(0),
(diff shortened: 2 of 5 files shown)


erudenko:
Closing this one, and I'd rather explain properly than leave it sitting open for another three months.

The concern is real. We have no index tracking in `anthropic-sse.ts` at all. But I don't think the clamp can be made correct, and that's the problem.

`clampIndex` remaps a `content_block_start` from index 2 to 1, but the block's deltas still arrive carrying index 2, and `clampIndex(2)` returns 2 unchanged, because `2 > highestSeenIndex + 1` is false. So the client gets deltas for a block it never opened, which is the same "Content block not found" this exists to prevent. The bound is also wrong for deltas. A delta can only point at a block that's already open, so the ceiling is `highestSeenIndex`, not plus one. As written it lets through exactly the one index that's guaranteed invalid.

Two other things. The `continue` in hunk 2 exits before main's usage and text tracking block, so `onAssistantText`, `textChunks` and `stopReason` get dropped for any repaired frame. And rewritten frames are enqueued directly instead of going through `enqueueData`, which skips the Layer 4 tool repair interceptor. So the frames most likely to need repair are the ones that stop getting it.

The underlying thing is that a stream index isn't a number you can clamp. It's an identity. The only shape that works is a `Map<upstreamIndex, emittedIndex>` filled on `content_block_start` and read on every delta and stop. That would also absorb the thinking filter's `thinkingBlocksSuppressed` subtraction, which is the same problem solved a third way in the same file.

If you have a `--debug` capture showing a real index jump, send it and I'll build that properly. Right now no fixture in the repo contains one, so I can't test a fix or prove the bug, and I'd rather not guess at this layer.

erudenko:
The bug is real and nothing on main handles it, so I want this. Two things before I can merge it.

**The clamp doesn't cover the delta that follows a remapped start.** Walk 0 → 2:

1. `content_block_start` idx 0. `expected` is 0, matches, passes through. `trackIndex(0)`, highest becomes 0.
2. `content_block_start` idx 2. `expected` is 1, so it's remapped to 1 and emitted as 1. `trackIndex(1)`, highest becomes 1.
3. `content_block_delta` idx 2. `clampIndex(2)` asks `2 > highest + 1`, that's `2 > 2`, false. Returns 2 unchanged and emits a delta for index **2**.

The client opened 0 and 1. It gets a delta for 2 and throws "Content block not found", which is the thing we're fixing. The `content_block_stop` at 2 lands the same way.

The condition is right for detecting a jump but wrong for following one. Once a start has been moved down, everything carrying that original index has to move with it. I think this wants a persistent `Map<originalIndex, emittedIndex>` populated at `content_block_start` and read by delta and stop, rather than a stateless clamp. Then step 3 looks up 2 and finds 1.

I hit the same shape in #172, where the thinking filter's renumbering path enqueues a different index than the frame carries. Anything that rewrites an index has to rewrite every frame that references it.

**The other thing is a fixture.** I went through every Anthropic-format capture we have, minimax turn1/2/3 and both z.ai ones, and the indices are strictly sequential in all of them:

```
minimax-m25-turn1:   0 0 0 0 1 1 1 2 2 2 2
regression-zai-glm5: 0 0 0 0
```

So I can't reproduce the gap, and I don't want to guess at behaviour I can't see. If you can catch one with `--debug` and drop the `[SSE:anthropic]` lines here, that becomes the regression test and I'll take the PR. `bun run packages/cli/src/test-fixtures/extract-sse-from-log.ts logs/claudish_*.log` pulls them straight out of the log.

Worth saying: rewriting indices in a byte-level passthrough is the riskiest thing we do in that file, so I'm being fussy about it rather than about you. Get me a capture and I'll help with the map.

Also fair warning, this file changes underneath you in #172. Happy to rebase it for you once that lands if you'd rather not.

erudenko:
Housekeeping note: I left a "Closing this one" comment on 9 August and then never actually closed it, so you've had a PR sitting open under a comment saying it was done. Sorry about that, it wasn't deliberate.

The review I posted today supersedes it. This one stays open and I'd like it landed.

jsboige:
Rebased onto current `main` (the previous head predated the `bd9bd85` rewrite of this parser). Scope notes:

- **Kept**: the index clamping/remapping itself — including a fix over the previous version. The old logic remapped a jumping `content_block_start` (z.ai 0 → 2) but not the block's *subsequent* frames: a following `content_block_delta` still carried the original index and passed the `highest + 1` check, so the client could still see an unknown index. The rebase carries a block-aware remap (original → remapped index held until `content_block_stop`).
- **Dropped**: the synthetic `message_stop` tail that was riding along in the old head — that is exactly what #172 covers, so it stays out of this PR.
- Noted in passing: `sseDataPayload` (c9e97c9) already covers the bare `data:` form we had a local patch for.

Added regression tests (`anthropic-sse-index-clamping.test.ts`): sequential pass-through unchanged, jumping start remapped with its follow-up frames, orphan delta clamped. All 28 stream-parser tests pass.

jsboige:
Your review aged well — the rebase pushed yesterday implements exactly the shape you asked for, plus the ceiling point from your 9 Aug comment folded in.

**The map.** `remappedBlocks: Map<upstreamIndex, emittedIndex>`, populated when a `content_block_start` gets remapped, read by every delta and stop, deleted at the block's `content_block_stop`. Your walkthrough now ends the way it should:

1. start idx 0 → passes, `highest` = 0
2. start idx 2 → `expected` 1, remapped and emitted as 1, `remappedBlocks.set(2, 1)`
3. delta idx 2 → map lookup returns 1, emitted as 1
4. stop idx 2 → map lookup returns 1, entry deleted

There's a regression test that walks 0 → 2 with the block's delta and stop and asserts the client never sees index 2 (this exact test fails against the previous stateless clamp — we hit the same step-3 leak independently while rebasing, which is what forced the map).

**The ceiling.** You were also right that the delta bound was wrong in the other direction: an orphan delta (no open block at that index) clamped to `highest + 1` — an index the client never opened, which is the same failure in a different costume. Deltas and stops now clamp to `highest` (only an already-opened block is valid); starts keep `highest + 1` (the next sequential slot is legitimate). The orphan-delta test asserts no `"index":1` escapes, which fails under the `+1` behavior.

**Your two other points** from the 9 Aug comment, also addressed while in there: rewritten frames go through `enqueueData` everywhere (Layer 4 tool repair preserved on the frames most likely to need it), and the `continue` that skipped usage/text tracking in the thinking-filter path is gone — every frame falls through to tracking, matching the passthrough path.

**The fixture — honest status.** We don't have one either. The original incident was a live z.ai GLM stream through our production proxy in May; the repo's captures postdate the fix, so they only show sequential indices (for the same reason yours do — our hub has been running a variant of this clamp since then). We're mining our raw-SSE production archives for a pre-fix capture; if one turns up we'll drop it as a `.sse` fixture via your extractor. Until then the three tests encode exactly the shapes you walked through, constructed to fail against both the leak and the `+1` clamp.

Noted on #172 landing under this — happy either way: rebase it yourself once yours merges, or ping us and we'll redo it on top.

jsboige:
**Update: we found your capture — 7 of them, and they taught us one more thing.**

We swept our production response archives (73 daily archives, ~200k captured streams). Every index fixture we had was sequential, same as yours — but the *client-facing* captures showed the leak itself, which reconstructs the upstream deterministically since the old clamp's semantics are known.

**7 hits, identical shape, all MiniMax-M3 on the anthropic passthrough** (Jul 19 → Aug 6):

```
content_block_delta  @0  signature_delta   ← NO content_block_start ever emitted
content_block_stop   @0                    ← (signature = sha256 of the empty string —
content_block_start  @0  {text}            ←  upstream start was @1, remapped down)
content_block_delta  @1  text_delta        ← every text delta LEAKS at the original index
```

So MiniMax runs the implicit-signature quirk you'd never guess from a fixture of clean streams: an entire block whose `start` never arrives. The old stateless clamp remapped the text start 1 → 0 and then, exactly per your walkthrough, passed every following delta through at index 1 — delta-for-a-block-the-client-never-opened. The client sees a delta before any start, a stop before any start, and text deltas with no home. Same family, second shape.

**What the new head does with that stream** (commit just pushed):

1. The orphan `signature_delta` and `stop` at index 0 — no remap entry, no open block — are **dropped**, event line swallowed with them. An orphan delta has no valid target: clamping it onto another open block would graft its content into that block. This is your "an index is an identity" taken to its conclusion — an identity with no referent doesn't get re-pointed, it gets refused.
2. The text `start` at 1 renumbers to 0 (highest was −1), `remappedBlocks` records 1 → 0.
3. Every text delta at 1 follows the map to 0.

Out the other end: one clean sequential block. The implicit signature block — which never existed client-side — vanishes entirely.

Two hardening notes: the clamp can no longer emit a negative index (the orphan-signature case previously clamped to `highest = −1`), and the thinking-filter path lost its last early-`continue` so repaired frames still hit usage/text tracking.

**On the fixture**: the two raw capture files from Aug 6 are being pulled out of cold storage to attach here — they'll become a `.sse` fixture via your extractor. The test in the new head (`anthropic-sse-index-clamping.test.ts`, "absorbs MiniMax's implicit signature block") already encodes the exact production sequence and fails against both the stateless clamp and the clamp-only intermediate we briefly had. Will link the real capture when it lands.

jsboige:
Saw #172 land this morning — rebased on top (clean, no conflicts). The two mechanisms are complementary: the clamp/orphan-drop only rewrites or drops `content_block_*` frames inside the block stream, and the terminal tail owns what happens after it ends; `finalizeAbandonedStream` is untouched and both test suites pass together (29/29 across the stream parsers).

The PR diff is now purely the index work — the synthetic `message_stop` tail I'd stripped from the original commits lives upstream where it belongs.

jsboige:
**Real fixtures delivered** — promised in my last reply, they took a detour (the captures live on a different machine than the branch; the shared-drive sync failed, they finally traveled over DM in base64).

Two production captures of MiniMax-M3 (anthropic passthrough), reqN=10324 and reqN=10416, now replayed in `anthropic-sse-index-clamping.test.ts`.

**One wrinkle worth documenting.** The on-disk captures tape the **outgoing** stream (the capture helper wraps `controller.enqueue`), i.e. the response *after* the old clamp had processed it. That turned out to be a gift: the old clamp remapped only `content_block_start` to sequential indices and let deltas/stops leak at their **upstream** index — so in the capture, every delta carries its true upstream index while every start carries the clamped one. Reconstructing the upstream stream is therefore mechanical (shift each start back to the index of the deltas that follow it), and it is **verifiable**: replaying the old clamp on the reconstructed stream reproduces the capture event-for-event (57/57 indexed events for r10324, 36/36 for r10416). The fixtures carry that proof in their headers.

The proven upstream shape is exactly the one from the synthetic test, now with the full tool-call tail:

- implicit signature block at 0: `signature_delta` (sha256 of the empty string) + `content_block_stop`, **no `content_block_start`**
- text block at 1 (start + ~30 deltas + stop, all at 1)
- `tool_use` block at 2 (start + `input_json_delta` chunks + stop, all at 2)

So under the old clamp the client received: a dangling stop at 0 (block never opened), text opened at 0 with every delta at 1, tool opened at 1 with every delta at 2 — "Content block not found" twice over.

The fixture tests assert, on both captures: the signature block is dropped whole, exactly two blocks are opened (text→0, tool_use→1), every delta follows its remap (zero leaks), the concatenated `partial_json` reassembles into valid tool input, stops close in order, and no emitted index ever escapes the sequential range.

One adjacent change: the `synthetic tail closes an open content block` test from #172 asserted the tail closes at the **upstream** index (3). Under renumbering the client has the block open at the clamped index, so the tail must close *that* — adapted, with a comment. Full neighborhood green: 128/128.

The fixture commit also re-derives the capture-side numbers end to end, so the whole chain (upstream shape → old-clamp behavior → client failure → new parser output) is now pinned by executable artifacts rather than prose.

erudenko:
Landed in #200. Thank you for this one.

The fixtures are what made it possible. I asked for a real capture and you came back with two, plus a reconstruction proof I hadn't asked for: replaying the old clamp against the reconstructed upstream and matching 57/57 and 36/36 indexed events. That's a stronger argument than most fixtures arrive with, and it's what let me verify the repair instead of taking it on trust.

I did add something before merging, and it's worth telling you about, because it's the kind of bug your own fixtures were pointing straight at.

The remapping only ran in one of the parser's two branches. `filterThinking` picks between them, and `MiniMaxModelDialect.shouldFilterThinking()` returns true unconditionally, so MiniMax, the provider your captures came from, took the branch without the fix. Measured on your r10324:

```
no adapter (what the tests pass)          opens [0, 1]   0 orphans
MiniMax-shaped adapter (what prod does)   opens [1, 2]   2 orphans
                                                         content_block_delta @ 0
                                                         content_block_stop  @ 0
```

Nothing ever opened index 0. So the turn still died, on the exact path the fixtures were captured from.

That isn't something you could have seen from the test file. Every test in there builds the parser with no `adapter`, and it's `composed-handler.ts` that passes the Layer 2 dialect in production. I only found it because I went looking for what the two branches did differently, which I did because you'd left them side by side and the asymmetry was visible.

The fix folds them into one `emitIndexed` layer. That let the thinking filter drop its suppressed-block counter too, since a suppressed block just never advances the index and the next block lands where it should. Your logic is the one that survived. It applies to both paths now.

Worth saying separately: the `pendingEventLine = null` on the dropped orphan is the same mechanism I asked for over on #133, and you'd applied it here before I got round to asking. That's why the orphan drop is clean rather than a clamp.

One thing that's on me. Sixteen fork CI runs were sitting queued and unapproved, some since July, so no test had ever run against any of your PRs. Fixed now, and it's most of why this sat as long as it did.

PR #128, opened 2026-05-16
feat(proxy): add /v1/models endpoint for gateway model discovery

jsboige:
## Summary

- Add `GET /v1/models` endpoint to the proxy server
- Returns all routable models from routing rules, profile model mappings, custom endpoints, and the `--model` flag
- Enables Claude Code's gateway model discovery (`CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1`)

## Details

The endpoint aggregates models from four sources:

1. **Routing rules** — model name patterns from `DEFAULT_ROUTING_RULES` + user config (catch-all `*` excluded)
2. **Profile model mappings** — opus/sonnet/haiku/subagent role mappings from `--model-map` or config
3. **Custom endpoints** — models declared in `customEndpoints[].models` from `~/.claudish/config.json`
4. **Default model** — the `--model` flag if set

Response format follows the OpenAI `/v1/models` convention:
```json
{
  "object": "list",
  "data": [
    { "id": "claude-sonnet-4-20250514", "display_name": "claude-sonnet-4-20250514", "created_at": "..." },
    ...
  ]
}
```

Zero new dependencies — uses only existing imports (`loadConfig`, `loadRoutingRules`). Config read failures for custom endpoints are silently skipped (proxy stays up).

## Test plan

- [ ] `bun run build` — compiles without errors
- [ ] `curl http://localhost:3000/v1/models` — returns model list
- [ ] With routing rules + custom endpoints configured, verify all models appear
- [ ] `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1` Claude Code session uses the endpoint correctly

### DIFF
--- a/packages/cli/src/proxy-server.ts
+++ b/packages/cli/src/proxy-server.ts
@@ -487,8 +487,6 @@ export async function createProxyServer(
   app.get("/v1/probe-discover", async (c) => {
     const provider = c.req.query("provider");
     if (!provider) return c.json({ error: "missing provider query" }, 400);
-    // Optional exclude list — TUI's probe loop passes models that already
-    // failed so discovery returns the next candidate. Format: comma-separated.
     const excludeQuery = c.req.query("exclude") ?? "";
     const exclude = new Set(
       excludeQuery
@@ -496,13 +494,7 @@ export async function createProxyServer(
         .map((s) => s.trim())
         .filter(Boolean)
     );
-    // Use a sentinel model name — the handler factory needs one, but
-    // discoverProbeModel doesn't consult the modelName field.
     const targetModel = `${provider}@<discover>`;
-    // Try local providers first (ollama, lmstudio, vllm, mlx). They're
-    // filtered out of the remote registry by design, so getRemoteProviderHandler
-    // returns null for them and we'd otherwise report "transport does not
-    // support discovery" even though LocalTransport DOES implement it.
     const handler =
       getLocalProviderHandler(targetModel) ?? getRemoteProviderHandler(targetModel);
     const transport = (handler as unknown as { provider?: ProviderTransport })?.provider;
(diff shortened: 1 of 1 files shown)


erudenko:
Closing this one. We shipped `/v1/models` in `f266cea`, about three weeks after you opened this, so it's on main now at `proxy-server.ts:748`.

The contract came out different and it's worth explaining, because it isn't obvious. Ours advertises Claude-recognised slot ids rather than raw model names. Claude Desktop silently drops any model id it doesn't recognise, so if you advertise the real names they just vanish from the picker with no error anywhere. That's also why non-serve proxies return an empty list on purpose rather than by accident.

Two things in your version I want to flag. Not as criticism, more because both would have failed quietly.

It enumerates `Object.keys(effectiveRoutingRules)` as model ids, and those keys are glob patterns. `gpt-*`, `gemini-*` and `*-zen` would all get advertised as selectable models. You filtered the `*` catch-all, so I think you half-saw it.

And GitHub reports this as mergeable, but Hono dispatches in registration order and ours registers first, so the new handler would never run at all. That one is nasty, because a duplicate route reads as working code in review.

The diff also strips 8 lines of comments off `/v1/probe-discover`, the exclude-list and sentinel-model rationale. I'm assuming that was accidental.

erudenko:
The route already exists on main, but not doing what you need, so this isn't a straight duplicate.

`proxy-server.ts:755` serves `/v1/models` for Claude Desktop's third-party inference mode. That picker builds itself only from a live `GET /v1/models` and silently drops any id it doesn't recognise, so the route deliberately advertises the Claude-recognised SLOT ids passed in via `options.servedSlotIds`, not the real model ids behind them. Narrow on purpose.

Which means merging this as-is would break Desktop. Your aggregate list is exactly the thing that picker discards.

But your case is real and currently unserved: `servedSlotIds` defaults to `[]` for every non-serve caller, so a normal `claudish --model X` session answers `/v1/models` with an empty list, and gateway discovery gets nothing.

I think these coexist cleanly:

```ts
const slots = options.servedSlotIds ?? [];
return c.json({
  object: "list",
  has_more: false,
  data: (slots.length > 0 ? slots : aggregateRoutableModels()).map(...),
});
```

Serve mode keeps its exact current output, byte for byte. Every other mode gets your four-source list instead of an empty array. No flag, no second route, and the two consumers never see each other's data.

If you're up for reshaping it that way I'll take it. Two things I'd want in the aggregate: the catch-all `*` rule stays excluded (you already do that), and models from providers with no credentials get filtered out, otherwise the picker offers things that can only 401.

One caveat I should flag: I don't run Claude Desktop against `serve` often enough to catch a regression there by feel, so the `slots.length > 0` branch needs to stay provably identical rather than merely equivalent-looking.

erudenko:
Housekeeping note: I left a "Closing this one" comment on 9 August and then never actually closed it, so you've had a PR sitting open under a comment saying it was done. Sorry about that, it wasn't deliberate.

The review I posted today supersedes it. This one stays open and I'd like it landed.

jsboige:
Reshaped and reposted as #180 following the coexistence pattern from your review. The original PR had two issues your review caught: route-shape (slots vs aggregate, kept as a single endpoint) and the catch-all '*' exposure. Closing this in favor of the replacement.

PR #131, opened 2026-05-28
fix(cli): detect Bun on Windows in launcher (which → where)

Trypd:
## Problem

On Windows, `claudish` fails to launch even when Bun is installed and on `PATH`:

```
'which' is not recognized as an internal or external command,
operable program or batch file.
claudish requires the Bun runtime but it was not found.
```

`bin/claudish.cjs` `findBun()` is POSIX-only:

1. `execSync("which bun")` — `which` doesn't exist in cmd/PowerShell, so the lookup throws.
2. The fallback candidates are all Unix paths (`$HOME/.bun/bin/bun`, `/usr/local/bin/bun`, `/opt/homebrew/bin/bun`) — none match the Windows install location `%USERPROFILE%\.bun\bin\bun.exe`.
3. `process.env.HOME` is typically unset on Windows (it's `USERPROFILE`), so even the first candidate resolves to `undefined/.bun/bin/bun`.

Net effect: a correctly installed Bun is reported as missing on every Windows machine.

## Fix

- Use `where bun` on `win32`, `which bun` elsewhere; take the first returned line.
- Add Windows fallback candidates `%USERPROFILE%\.bun\bin\bun.exe` (and extensionless), using `HOME || USERPROFILE`.
- Make the "install Bun" hint platform-aware — Windows users get the PowerShell one-liner (`irm bun.sh/install.ps1 | iex`) instead of `curl … | bash`.

No behavior change on macOS/Linux (same `which` path and same Unix candidates).

## Testing

- Windows 11 + Bun 1.3.13 on PATH: `claudish --version` → `claudish version 7.1.2` (was failing before).
- `node --check bin/claudish.cjs` passes.

### DIFF
--- a/packages/cli/bin/claudish.cjs
+++ b/packages/cli/bin/claudish.cjs
@@ -8,17 +8,27 @@ const { execFileSync, execSync } = require("child_process");
 const { resolve } = require("path");
 
 function findBun() {
+  const isWin = process.platform === "win32";
+  // PATH lookup — "which" is POSIX-only; Windows uses "where".
   try {
-    const path = execSync("which bun", { encoding: "utf-8" }).trim();
-    if (path) return path;
+    const out = execSync(isWin ? "where bun" : "which bun", { encoding: "utf-8" }).trim();
+    const first = out.split(/\r?\n/).map((l) => l.trim()).find(Boolean);
+    if (first) return first;
   } catch {}
   // Common install locations
-  const candidates = [
-    process.env.HOME + "/.bun/bin/bun",
-    "/usr/local/bin/bun",
-    "/opt/homebrew/bin/bun",
-  ];
+  const home = process.env.HOME || process.env.USERPROFILE;
+  const candidates = isWin
+    ? [
+        home && resolve(home, ".bun", "bin", "bun.exe"),
+        home && resolve(home, ".bun", "bin", "bun"),
+      ]
+    : [
+        home && home + "/.bun/bin/bun",
+        "/usr/local/bin/bun",
+        "/opt/homebrew/bin/bun",
+      ];
   for (const c of candidates) {
+    if (!c) continue;
     try {
       execFileSync(c, ["--version"], { stdio: "ignore" });
       return c;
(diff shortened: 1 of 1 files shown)


erudenko:
Good catch, and this has been broken longer than it should have been. `claude-runner.ts:1533` already uses `where` on Windows, so we treat Windows as supported everywhere except this one file. Nobody noticed because the error says "Bun runtime not found", which sends you looking at your Bun install instead of at us.

The `HOME || USERPROFILE` change fixes a second bug you might not have been aiming at. On POSIX with `HOME` unset we were building the literal string "undefined/.bun/bin/bun" and then testing whether it exists. Your `home &&` guard kills that too.

One thing worth adding while this is open. The spawn at `claudish.cjs:47` is untouched, and it's `spawnSync(bun, ...)` with no shell. If `where bun` returns a `.cmd` shim first, which is what happens with an npm-installed Bun, Node 18 and up refuses to exec it. We already hit this elsewhere, `claude-runner.ts:1306` has `needsShell = isWindows() && claudeBinary.endsWith(".cmd")`. The same guard would work here.

Which install did you test against, the official installer or npm? That decides whether the `.cmd` case is theoretical or the one most Windows users would actually land on.

The rebase is cosmetic, `bc2ead4` reformatted those exact lines to `node:child_process` imports. Sort that and the shell guard and I'll merge it.

erudenko:
Picked this up as #178, credited to you. Closing here since the branch had gone conflicting.

Your diagnosis was exactly right and I kept both halves of it, the `where` swap and the Windows candidate paths. Two things I added on top.

`where` prints every match on PATH, one per line, so the first line gets taken rather than the whole buffer. And the install hint is now platform-aware, because a Windows user who gets past the first dead end and reads `curl -fsSL https://bun.sh/install | bash` in cmd just hits a second one.

The part I want to flag: this survived because `bin/claudish.cjs` had no tests, and nothing in it was reachable without actually being on Windows. So the platform decisions are now three pure helpers and the file exports them when required instead of run, guarded on `require.main === module`. Six hermetic tests over those. The one I like most asserts `bunCandidates("win32", {})` contains no "undefined", since HOME being unset is the case a real Windows user hits and the old code turned that into the literal path `undefined\.bun\bin\bun.exe`.

I can't run Windows here, so that branch is covered by unit tests rather than an end to end launch. If you get a chance to try #178 on your machine I'd genuinely value hearing whether it works.

PR #133, opened 2026-05-30
fix(anthropic-sse): suppress server_tool_use blocks and handle empty responses

jsboige:
## Problem

Z.AI and GLM providers emit `server_tool_use` content blocks for their built-in tools (e.g., web search). These blocks confuse downstream consumers (Claude Code) that only expect `text`, `thinking`, and `tool_use` block types, causing parse errors or dropped content.

Additionally, some providers occasionally return completely empty responses (no `message_start` event at all), which causes the client to hang waiting for content.

## Changes

**File:** `packages/cli/src/handlers/shared/stream-parsers/anthropic-sse.ts`

1. **Server tool suppression** -- Detect `content_block_start` with `type: "server_tool_use"` in both the filterThinking and non-filterThinking code paths and suppress them (skip the SSE event). A counter tracks nesting so `content_block_stop` events are properly matched.

2. **Message lifecycle tracking** -- Track `sawMessageStart` and `sawMessageStop` flags so the finalizer can detect when the provider omitted lifecycle events.

3. **Empty response handling** -- When no `message_start` was received, emit a synthetic message with a text block containing an actionable error message ("Try compacting the conversation or reducing the context size").

4. **Synthetic message_stop** -- When the provider omits `message_delta`+`message_stop`, emit them so the client stream terminates cleanly.

## Providers affected

- Z.AI (GLM models via Anthropic transport)
- GLM direct
- MiniMax, Kimi ( Anthropic transport, rare edge case)

## How to test

```bash
claudish --model zai@glm-4.7 --debug "search the web for Claude Code"
# Verify: no server_tool_use blocks in output, stream terminates normally

claudish --model zai@glm-4.7 --debug ""  # empty prompt, may trigger empty response
# Verify: synthetic error message appears, no hang
```

### DIFF
--- a/packages/cli/src/handlers/shared/stream-parsers/anthropic-sse.ts
+++ b/packages/cli/src/handlers/shared/stream-parsers/anthropic-sse.ts
@@ -68,6 +68,9 @@ export function createAnthropicPassthroughStream(
           let textChunks = 0;
           let toolUseBlocks = 0;
           let stopReason: string | null = null;
+          let sawMessageStart = false;
+          let sawMessageStop = false;
+          let suppressedServerTools = 0;
 
           // Thinking-block filtering state
           let insideThinkingBlock = false;
@@ -137,6 +140,20 @@ export function createAnthropicPassthroughStream(
                     continue;
                   }
 
+                  // Suppress server_tool_use blocks (Z.AI built-in tools)
+                  if (
+                    data.type === "content_block_start" &&
+                    data.content_block?.type === "server_tool_use"
+                  ) {
+                    suppressedServerTools++;
+                    log(`[AnthropicSSE] Suppressing server_tool_use block at index ${data.index}`);
+                    continue;
+                  }
+                  if (data.type === "content_block_stop" && suppressedServerTools > 0) {
+                    suppressedServerTools--;
+                    continue;
+                  }
+
                   // Re-index non-thinking content blocks
                   // After suppressing N thinking blocks, subtract N from the index
                   if (typeof data.index === "number" && thinkingBlocksSuppressed > 0) {
@@ -189,6 +206,20 @@ export function createAnthropicPassthroughStream(
                       return; // stop processing further lines
                     }
 
+                    // Suppress server_tool_use blocks (Z.AI built-in tools)
+                    if (
+                      data.type === "content_block_start" &&
+                      data.content_block?.type === "server_tool_use"
+                    ) {
+                      suppressedServerTools++;
+                      log(`[AnthropicSSE] Suppressing server_tool_use block at index ${data.index}`);
+                      continue;
+                    }
+                    if (data.type === "content_block_stop" && suppressedServerTools > 0) {
+                      suppressedServerTools--;
+                      continue;
+                    }
+
                     // No error — pass through the line
                     if (!isClosed) {
                       controller.enqueue(encoder.encode(line + "\n"));
@@ -219,6 +250,12 @@ export function createAnthropicPassthroughStream(
                     }
                     if (data.type === "message_delta" && data.delta?.stop_reason) {
                       stopReason = data.delta.stop_reason;
+                    }
+                    if (data.type === "message_stop") {
+                      sawMessageStop = true;
+                    }
+                    if (data.type === "message_start") {
+                      sawMessageStart = true;
                     }
                   } catch {
                     // Unparseable data line — pass through
@@ -261,6 +298,12 @@ export function createAnthropicPassthroughStream(
                   if (data.type === "message_delta" && data.delta?.stop_reason) {
                     stopReason = data.delta.stop_reason;
                   }
+                  if (data.type === "message_stop") {
+                    sawMessageStop = true;
+                  }
+                  if (data.type === "message_start") {
+                    sawMessageStart = true;
+                  }
                 } catch {}
               }
             }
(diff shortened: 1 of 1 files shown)


erudenko:
There are two different changes in here and I think they need splitting.

The empty-response half is worth having. We do just close at `anthropic-sse.ts:464` with nothing synthesized, so a turn that arrives with no `message_start` hangs the client. But that's the same fix as #143 and #127, three copies of one missing helper. I'm building a single `finalizeStream` for it, and that's where this half will land.

The suppression half I can't take as written, because I think it makes things worse rather than better.

It drops the `content_block_start` and a `content_block_stop`, but the block's `content_block_delta` frames still go through carrying an index the client never opened. That's the exact "Content block not found" #127 exists to fix. Compare the thinking filter right above it, which needs a full `insideThinkingBlock` gate to suppress the deltas as well.

Indices aren't renumbered either. The thinking path subtracts `thinkingBlocksSuppressed` at `anthropic-sse.ts:310`, this doesn't, so suppressing block N leaves a permanent gap.

And the `content_block_stop` counter ignores `data.index`, so it swallows whichever stop arrives next rather than the matching one.

There's a rebase hazard on top of that. Main enqueues non-data lines eagerly now, so a bare `continue` on the unfiltered path leaves `event: content_block_start` with no body behind it. That's the exact failure the `pendingEventLine` machinery was built to prevent.

What I'd need to take this: a captured `.sse` fixture showing Z.AI actually emitting `server_tool_use`. Nothing in `test-fixtures` has one, so right now I can't test it or even confirm the frame shape. Then an `insideServerToolBlock` gate that mirrors the thinking one, deltas included, with renumbering, and suppression routed through `pendingEventLine = null; suppressedFrame = true` instead of `continue`.

If you still have the debug log from the run that prompted this, that's the fastest way to move it forward.

erudenko:
Both halves are real problems and nothing on main handles either. The suppression needs rework before I can take it.

**The stop counter isn't matched to a block.**

```ts
if (data.type === "content_block_stop" && suppressedServerTools > 0) {
  suppressedServerTools--;
  continue;
}
```

That drops the *next* `content_block_stop` that arrives, not the stop belonging to the suppressed block. If a `server_tool_use` opens at index 1 while the text block at index 0 is still open, the text block's stop is the one swallowed. Index 0 then never closes.

**And the suppressed block's deltas still go out.** Only `content_block_start` and one stop are filtered, so `input_json_delta` frames for that block reach the client referencing an index it never opened. That's "Content block not found" again.

Both want the same thing: track the suppressed block's index in a `Set`, drop every frame whose index is in it, and remove the index when its own stop arrives. Then it doesn't matter what order blocks interleave in.

**The part I'd want you to see: this PR and #127 are working against each other.** Suppressing a block without renumbering leaves a gap, so a turn with `server_tool_use` at index 1 emits 0, 2, 3. That is precisely the non-sequential-index condition #127 exists to clamp. Land both as-is and #133 manufactures the bug #127 patches, on the same file.

The thinking filter already solved this shape: it renumbers by subtracting the suppressed count (`anthropic-sse.ts:310`). Server tools should go through the same path rather than a parallel one. Honestly I think #133 and #127 want to be one PR with a single index-mapping layer that both the thinking filter and the server-tool filter feed.

**On the empty-response half** — worth knowing #172 (open now) adds a synthetic tail when a stream is abandoned, including a `message_start` when upstream never sent one. It fires on the error path though, not on a clean close with zero events, so your case probably survives it. Rebase on that and we'll see what's left.

**And I need a capture.** I have both `regression-zai-*` fixtures and neither contains a `server_tool_use`, so I can't reproduce this or write a test from real data. `--debug` then `bun run packages/cli/src/test-fixtures/extract-sse-from-log.ts logs/claudish_*.log` gets it out.

Genuinely good catches, both of them. This file is just the one place where getting it 90% right still breaks the turn.

erudenko:
Both halves of this have now been answered, but neither by this PR, so I'm closing it and want to say exactly where each one went.

**The empty-response half is on main.** #172 landed a synthetic terminal tail, including a `message_start` when upstream never sent one, so a stream that arrives with nothing no longer hangs the client. That was the half I said was worth having, and it is.

**The suppression half is superseded by a different shape.** #200 just landed @jsboige's index work from #127, and one thing I added on top matters directly to you: both of the parser's branches now go through a single `emitIndexed` layer that owns index mapping. Remap a start that jumped, follow that mapping for the block's own deltas and stop, drop an orphan whole, delete the mapping when the block closes.

That layer is where server-tool suppression belongs, and it removes the three problems I raised:

- the stop counter can't mismatch a block, because mapping is keyed by index rather than counted
- a suppressed block's deltas can't leak, because they resolve through the same map
- indices can't gap, because a suppressed block simply never advances `highestSeenIndex` and the next real block takes the slot

That last one is worth dwelling on. The old thinking filter renumbered by subtracting a suppressed count, and your PR would have added a second, independent suppression without renumbering. Landed together, #133 would have manufactured exactly the gap #127 was patching, in the same file. That's why I wanted them as one change and not two.

**What still blocks it: I have no capture.** Both `regression-zai-*` fixtures are on disk and neither contains a `server_tool_use`, so I can't reproduce this, write a test from real data, or even confirm the frame shape you're filtering. That constraint hasn't moved since June, and I'm not going to merge a stream filter written against a shape nobody has recorded.

If you still hit it, this is the whole ask:

```
claudish --model <the z.ai model> --debug   # reproduce
bun run packages/cli/src/test-fixtures/extract-sse-from-log.ts logs/claudish_*.log
```

Send me that `.sse` and I'll build the suppression into `emitIndexed` myself, with credit to you. It'd be a small function at that point rather than a parallel filter, which is the outcome I was hoping for when I first pushed back.

Closing on the state of the diff, not on the finding. Both problems you reported were real and one of them is fixed.

PR #134, opened 2026-05-30
fix(composed-handler): strip inline system messages for Anthropic-transport providers

jsboige:
## Problem

Claude Code v2.1.153+ injects `role: "system"` messages inline in the messages array (e.g., system-reminders injected between turns). Anthropic-compat providers like Z.AI and GLM reject these with HTTP 400 because they only accept `user` and `assistant` roles in messages.

### `packages/cli/src/handlers/composed-handler.ts`

After step 4 (buildPayload), add a new step 4b that detects `anthropic-sse` transport format and strips any `role: "system"` messages from the payload's messages array. Their text content is merged into the top-level `system` prompt field instead (appended to any existing system prompt).

### `packages/cli/src/handlers/shared/format/openai-messages.ts`

Add a handler for `msg.role === "system"` in the message processing loop. System messages are merged into the existing system prompt (prepended or appended to the first system message) rather than silently dropped or causing a protocol error.

## Providers affected

- Z.AI (Anthropic transport)
- GLM direct (Anthropic transport)
- MiniMax, Kimi (Anthropic transport)
- Any future Anthropic-compat provider

## How to test

1. Use Claude Code v2.1.153+ which injects system-reminders
2. Route through claudish with an Anthropic-transport provider:
   ```bash
   claudish --model zai@glm-4.7 "hello"
   ```
3. Without this fix: HTTP 400 from Z.AI ("invalid role: system")
4. With this fix: System reminders silently merged into system prompt, request succeeds

### DIFF
--- a/packages/cli/src/handlers/composed-handler.ts
+++ b/packages/cli/src/handlers/composed-handler.ts
@@ -301,6 +301,38 @@ export class ComposedHandler implements ModelHandler {
     // 4. Build request payload
     let requestPayload = adapter.buildPayload(claudeRequest, messages, tools);
 
+    // 4b. Strip inline system messages from messages[] for Anthropic-transport providers.
+    // Claude Code v2.1.153+ injects system messages inline (e.g. system-reminders).
+    // Providers like Z.AI reject role:"system" in messages — only role:"user"/"assistant" accepted.
+    // Merge them into the top-level system field instead.
+    if (this.provider.streamFormat === "anthropic-sse" && requestPayload.messages) {
+      const inlineSystemTexts: string[] = [];
+      requestPayload.messages = requestPayload.messages.filter((msg: any) => {
+        if (msg.role === "system") {
+          const text = typeof msg.content === "string"
+            ? msg.content
+            : Array.isArray(msg.content)
+              ? msg.content.map((c: any) => c.text || "").join("\n")
+              : "";
+          if (text) inlineSystemTexts.push(text);
+          return false;
+        }
+        return true;
+      });
+      if (inlineSystemTexts.length > 0) {
+        const merged = inlineSystemTexts.join("\n\n");
+        if (requestPayload.system) {
+          const existing = Array.isArray(requestPayload.system)
+            ? requestPayload.system.map((s: any) => s.text || s).join("\n\n")
+            : requestPayload.system;
+          requestPayload.system = existing + "\n\n" + merged;
+        } else {
+          requestPayload.system = merged;
+        }
+        log(`[ComposedHandler] Merged ${inlineSystemTexts.length} inline system message(s) into system prompt for ${this.provider.displayName}`);
+      }
+    }
+
     // Merge provider-specific extra fields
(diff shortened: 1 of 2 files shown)


erudenko:
The gap here is real and I hadn't spotted it. `openai-messages.ts:38` only handles user and assistant, so a `role:"system"` message gets silently dropped, and `anthropic-api-format.ts:39` passes it straight through to providers that reject it. Two different bad outcomes out of the same hole.

Something related I found while checking. `transform.ts:184` `transformMessages()` would have hoisted system messages, and it has zero callers. So the fix already exists in the tree, just disconnected from everything. Worth knowing.

Before I merge though, has Claude Code actually sent you one? The only captured request we have has user roles only, and the Messages API schema CC speaks doesn't have a system role inside `messages` at all. If you hit it for real, a `--debug` capture settles it. If it's defensive hardening that's fine too, I'd just want the commit message to say which.

Two things I'd change either way.

Position gets destroyed. A system-reminder injected between turn N and N+1 ends up appended to the end of the global system prompt, so it loses the "this applies right now" placement that made it useful. The OpenAI-path hunk has the mirror problem, it unshifts to the front of `messages`. Same input, two different reorderings depending on which provider you're on.

And it's two implementations of one merge in two files. That belongs in `transform.ts`, ideally by reviving `transformMessages` since it's sitting right there doing nothing. One place, both paths fixed, dead code gone.

`this.provider.streamFormat` is the right gate today, but it reads the raw field rather than `overrideStreamFormat()`. Harmless now because OpenRouter and LiteLLM both override toward openai-sse, but it'll bite whenever one overrides the other way.

erudenko:
Confirmed and picked up as #179, credited to you. Closing this since I moved where it lives.

I verified the entry really does survive to the wire:

```
transformOpenAIToClaude({ messages: [user, system, assistant] })
-> roles left in messages: user, system, assistant
```

The thing that made this worth digging into: the fix already existed and was unreachable. `transform.ts` has `transformMessages()`, which hoists exactly these entries into `req.system` and assigns the cleaned array back. It's exported and called from nowhere. Dead since before the adapter layer existed. It also overwrites `req.system` rather than appending, so anyone who found it and wired it up would have dropped the user's real system prompt.

Two changes from your version.

I put it in `AnthropicAPIFormat.buildPayload` instead of a step 4b in ComposedHandler. It's a wire-format legality rule and that class is the Anthropic wire format, so it doesn't need to know which providers use anthropic transport. Your gate was `this.provider.streamFormat === "anthropic-sse"`, and the resolved format actually comes from a three-tier lookup (`transport.overrideStreamFormat() ?? modelAdapter.getStreamFormat() ?? providerAdapter.getStreamFormat()`), so reading the raw field would have missed providers.

And the system array stays an array. Yours joins blocks into a string when merging. Claude Code puts `cache_control` on those blocks, so flattening them drops the cache breakpoints, which quietly turns a 400 fix into a bigger bill. There's a test pinning that now.

Good find. This is one of those bugs where the error message points at the provider and the cause is three layers upstream.

PR #135, opened 2026-05-30
feat(web-search): intercept web search tool calls and execute via SearXNG

jsboige:
## Problem

When providers (Z.AI, GLM, OpenRouter models) emit WebSearch/WebFetch tool calls, they fail because the proxy doesn't implement server-side web search. GLM models also emit `<searchWeb>query</searchWeb>` tags in their text output, which go unhandled. Additionally, Claude Code's built-in WebSearch tool fires sub-agent requests that get forwarded to the model provider instead of being executed locally.

### New file: `packages/cli/src/handlers/shared/web-search-executor.ts`

A stream-safe SearXNG client that:
- Queries a local SearXNG instance (`SEARXNG_URL` env var, defaults to `http://search.myia.io`)
- Races SearXNG against a configurable deadline (default 3s)
- Returns formatted results as text, never throws
- Extracts search queries from tool call arguments JSON

### `packages/cli/src/proxy-server.ts`

Add `interceptWebTools()` function that intercepts sub-agent web search/fetch requests BEFORE handler selection. Detects the standard Claude Code WebSearch sub-agent pattern ("Perform a web search for the query: X") and returns SearXNG results as a streaming or non-streaming Anthropic-format response.

### `packages/cli/src/handlers/shared/stream-parsers/openai-sse.ts`

Three enhancements:

1. **WebSearch tool suppression** -- When a structured WebSearch/WebFetch tool call is detected, set `suppressed = true` and buffer arguments. At finalization, execute SearXNG and inject results as a text block.

2. **GLM `<searchWeb>` tag detection** -- Detect `<searchWeb>query</searchWeb>` tags in accumulated text and execute SearXNG for each query.

3. **Empty response handling** -- When no text or tool blocks were started, inject a synthetic error message so downstream doesn't hang.

## Providers affected

- Z.AI (GLM models)
- GLM direct
- OpenRouter models that emit web search tool calls
- Any provider where Claude Code's WebSearch sub-agent is routed through claudish

## How to test

1. **Sub-agent interception:**
   ```bash
   claudish --model zai@glm-4.7 "search the web for recent AI news"
   # Verify: SearXNG results appear in response instead of tool call error
   ```

2. **GLM <searchWeb> tags:**
   ```bash
   claudish --model glm-4.7 --debug "search for..."
   # Check debug log for [Streaming] Found N <searchWeb> tag(s) in text
   ```

3. **SearXNG URL configuration:**
   ```bash
   SEARXNG_URL=http://localhost:8080 claudish --model glm-4.7 "search for..."
   ```

## Notes

- SearXNG is expected to be running and accessible. If SearXNG is unavailable, a timeout/failure message is injected instead of crashing.
- The `SEARXNG_URL` env var should be documented in the main README as an optional configuration.

### DIFF
--- a/packages/cli/src/handlers/shared/stream-parsers/openai-sse.ts
+++ b/packages/cli/src/handlers/shared/stream-parsers/openai-sse.ts
@@ -16,6 +16,7 @@ import {
   type ToolSchema,
 } from "../tool-call-recovery.js";
 import { isWebSearchToolCall, warnWebSearchUnsupported } from "../web-search-detector.js";
+import { executeWebSearch, extractSearchQuery } from "../web-search-executor.js";
 
 export interface StreamingState {
   usage: any;
@@ -39,6 +40,7 @@ export interface ToolState {
   closed: boolean;
   arguments: string; // Accumulated JSON arguments string
   buffered: boolean; // Whether we're buffering args until tool call completes
+  suppressed?: boolean; // Whether this tool call was suppressed (e.g. web search intercepted)
 }
 
 /**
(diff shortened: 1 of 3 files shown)


jsboige:
Superseded by the consolidated PR below — all web-search interception work is included there.

PR #136, opened 2026-05-30
Add Tuning Engines provider

cerebrixos:
Adds Tuning Engines as an OpenAI-compatible provider option.

This includes the Tuning Engines API base URL, key placeholder, and starter model IDs where this project exposes provider presets, examples, or configuration docs.

Validation:
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 **Claudish** (Claude-ish) is a CLI tool that allows you to run Claude Code with any AI model by proxying requests through a local Anthropic API-compatible server.
 
 **Supported Providers:**
-- **Cloud:** OpenRouter (580+ models), Google Gemini, OpenAI, MiniMax, Kimi, GLM, Z.AI, OllamaCloud, OpenCode Zen
+- **Cloud:** OpenRouter (580+ models), Google Gemini, OpenAI, Tuning Engines, MiniMax, Kimi, GLM, Z.AI, OllamaCloud, OpenCode Zen
 - **Local:** Ollama, LM Studio, vLLM, MLX
 - **Enterprise:** Vertex AI (Google Cloud)
 
@@ -322,6 +322,7 @@ Claudish automatically loads `.env` from the current directory at startup. For t
 | `OPENROUTER_API_KEY` | OpenRouter (default backend, 580+ models) | |
 | `GEMINI_API_KEY` | Google Gemini (`g@`, `google@`) | |
 | `OPENAI_API_KEY` | OpenAI (`oai@`) | |
+| `TUNING_ENGINES_API_KEY` | Tuning Engines (`te@`, `tuningengines@`) | |
 | `MINIMAX_API_KEY` | MiniMax (`mm@`, `mmax@`) | |
 | `MINIMAX_CODING_API_KEY` | MiniMax Coding Plan (`mmc@`) | |
 | `MOONSHOT_API_KEY` | Kimi/Moonshot (`kimi@`) | `KIMI_API_KEY` |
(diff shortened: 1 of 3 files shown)


cerebrixos:
Hello team @svallory  . Please do consider this PR for merge. So much appreciate in advance :)

erudenko:
The wiring is right, which is the part people usually get wrong. Both tables updated, `BUILTIN_PROVIDERS` and `PROVIDER_PROFILES`, so this doesn't hit the silent OpenRouter fallback we've been bitten by before. The `te` and `tuning` shortcuts don't collide with anything across the other 32 providers. Credential wiring is derived, so there's no third table to remember. I checked the endpoint as well, `api.tuningengines.com` resolves and `/v1/models` gives a 401 rather than a connection error, so the gateway is live.

Three things before I merge.

There's a trap you had no way of knowing about. `openaiProfile.createHandler` calls `requiresResponsesApi(ctx.modelName)` at `provider-profiles.ts:155` and silently rewrites apiPath to `/v1/responses` for gpt-5.x-shaped names. If Tuning Engines re-serves any model named like that, requests go to an endpoint their gateway probably doesn't implement, and it'll look like their bug rather than ours. Either confirm they don't serve those, or this needs its own profile instead of `openaiProfile`.

`nativeModelPatterns: [/^tuningengines\//i, /^te\//i]` claims `te/` globally, so any bare `te/foo` auto-detects here. That's a very short namespace to take across the whole tool. The `legacyPrefixes` entries already cover explicit use, so I'd drop the `te/` pattern and keep the long one.

And nothing verifies any of it. No test, no probe or smoke entry, no CHANGELOG. The README example `te@llama-3.3-70b-fp8` is unverified. Your description mentions starter model ids but the diff has none, which is actually correct under our no-hardcoded-roster rule, it just means nothing exercises the wiring at all. One live-verified model id in a smoke entry would be enough.

Do you use this gateway yourself? Changes how much weight I put on that last point.

erudenko:
Technically this is one of the cleanest provider additions I've seen on this repo, so the hold-up isn't the code.

What I checked:

- Both tables. `BUILTIN_PROVIDERS` **and** `tuningengines: openaiProfile`. A definition without a profile silently falls through to OpenRouter, which means someone's requests quietly go to the wrong vendor and get billed there. You got that right, and most people don't.
- No hardcoded model roster, which this codebase is strict about.
- `te`, `tuning`, `tuningengines` are all free across the 32 existing providers.
- Endpoint is live: `GET https://api.tuningengines.com/v1/models` answers 401, so it's real and wants auth.

One thing I'd change: `nativeModelPatterns: [{ pattern: /^te\//i }]` claims a two-letter namespace for BARE model names, so anyone typing a model called `te/whatever` gets routed here regardless of intent. The `@` prefix already covers explicit use. I'd drop that pattern and keep `/^tuningengines\//i`.

The real question is policy, not code, and it's not mine alone to answer. Claudish has custom endpoints for exactly this shape:

```json
{ "customEndpoints": {
    "tuning": { "kind": "simple", "url": "https://api.tuningengines.com",
                "format": "openai", "apiKey": "${TUNING_ENGINES_API_KEY}" } } }
```

That gets you `tuning@model` today with no code change and no release. A builtin entry earns its place when a provider needs bespoke transport, auth, or quota handling — an OpenAI-compatible gateway generally doesn't. At 32 builtins the marginal one is mostly maintenance: probe coverage, the config TUI, the picker, and someone has to notice when the base URL moves.

So I'm leaving this open for @erudenko rather than deciding it myself. If the answer is yes, drop the `/^te\//i` pattern and I'll take it as-is.

Either way, thanks for doing it properly. The two-table thing in particular.

erudenko:
Your vendor shipped in v7.51.0, just not as a built-in provider.

I ended up building a bundled catalog of predefined custom endpoints (#185), and tuningengines is one of the 25 rows. Same `https://api.tuningengines.com`, same `TUNING_ENGINES_API_KEY`. Before adding it I probed the endpoint independently, with a deliberately invalid key against the configured path and a bogus sibling path, and got a clean 401 from your route. Your values were right.

The part I want to call out is `TUNING_ENGINES_BASE_URL`. I hadn't planned a base URL override at all. Reading your diff is what made me add it as a requirement, because a self-hosted gateway doesn't live at the public hostname and without the override those users can't touch the bundled row. It's in the schema now as `baseUrlEnvVars`, and a malformed override skips the entry rather than quietly falling back to `api.tuningengines.com`. That last bit matters more than it sounds: falling back would send a private instance's prompts and a key minted for it to a public host.

So I'm closing this, and it's about the code path, not the work. A catalog row is one object literal in `providers/predefined-catalog.ts`. No `ProviderDefinition`, no `PROVIDER_PROFILES` entry, no transport. That distinction is load-bearing in this repo, because a definition without a matching profile silently falls through to OpenRouter and bills the user there with no error. A data row can't create that failure, and adding vendor 26 needs no code review of routing at all.

One change to flag: the prefix is `tuningengines@` only. `te@` is gone, since a catalog row gets its own name as its single shortcut and I didn't want two-letter prefixes competing in the partial matcher.

Thanks for this. The base URL override was the good idea I'd have missed.

PR #137, opened 2026-05-31
fix: stream hang prevention, thinking block strip, response capture, web search interception

jsboige:
## Summary

Multiple bug fixes and features accumulated on the fork, all backwards-compatible:

### Bug fixes
- **fix(stream-parsers): prevent openai-sse stream hang** (34bac1f) — `finalize()` referenced undefined `opts.modelName`, causing a `ReferenceError` before `controller.close()`. The error was swallowed, leaving the ReadableStream permanently open and the client hanging forever. Fixed by using the correct `target` parameter.
- **fix(composed-handler): strip thinking blocks from message history for non-native providers** (e1c9753) — Opus thinking blocks (with Anthropic signatures) were forwarded verbatim to Z.AI/GLM/MiniMax, causing corruption or API rejection.
- **fix(anthropic-sse): suppress server_tool_use blocks and handle empty responses** (0cb986a) — Z.AI sends `server_tool_use` blocks that the Anthropic client doesn't understand. Also emits synthetic `message_start`/`message_stop` when the provider returns an empty response.
- **fix(composed-handler): strip inline system messages for Anthropic-transport providers** (032919c) — Claude Code v2.1.153+ injects `role: "system"` inline. Anthropic-compat providers reject these.

### Features
- **feat(stream-parsers): intercept WebSearch/WebFetch via SearXNG** (24ec4da) — Replaces the \"web search unsupported\" warning with actual execution via a local SearXNG instance. Two interception paths: structured tool_call and GLM `<searchWeb>` tags.
- **feat(request-logger): add full-body capture mode for diagnostics** (122b681) — Gated by `CLAUDISH_CAPTURE_DIR` env var. No-op in production.

### Infrastructure
- Response-side SSE capture (`response-capture.ts`) — diagnostic helper gated by `CLAUDISH_CAPTURE_DIR`, wired into all three stream parsers
- Docker: bump bun 1.3-alpine, use SearXNG DNS name

## Testing
- 2000+ response captures during live testing: 0 `closed=false` (all streams close cleanly)
- Tested 3 sub-agent types (code-explorer, Explore, general-purpose) under GLM — all completed
- Tested Opus sub-agent with 35 tool_use blocks — completed in 8.8s
- All existing tests pass (`bun test`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #140, opened 2026-06-02
fix(native-handler): strip unsigned thinking blocks to fix mixed-session 400

jsboige:
## Summary

Fixes a `400 messages.N.content.M: Invalid signature in thinking block` error that hits **mixed-provider** Claude Code sessions when a turn routes back to a real Anthropic model.

### The bug

Non-Anthropic providers (GLM/Z.AI, Kimi, DeepSeek) emit reasoning that the `openai-sse` / `gemini-sse` stream parsers surface to the client as Anthropic `{ type: "thinking" }` content blocks — but **with no signature** (Claude Code stores them as `signature: ""`).

While the session stays on a non-native provider this is harmless: `ComposedHandler` already strips all thinking blocks before forwarding (see #137, the *forward* direction). The blocks become toxic in a **mixed** session — once a turn routes to a native Anthropic model, `NativeHandler` is a transparent passthrough and forwards the accumulated history verbatim to `api.anthropic.com`. The unsigned blocks are then rejected:

```
API Error: 400 messages.19.content.0: Invalid signature in thinking block
```

(An empty signature is "present but invalid", which is why this surfaces as *invalid* rather than *missing*.)

### The fix

A small pure helper, `stripUnsignedThinkingBlocks()`, removes **only** thinking blocks whose signature is missing or empty, called in `NativeHandler` just before the upstream fetch.

- **Genuine Anthropic thinking is preserved** — it carries a real signature and is required for interleaved-thinking / tool-use continuity.
- Other block types (`text`, `tool_use`, `tool_result`, `redacted_thinking`) are untouched.
- User messages are never modified.
- No-op for pure-native sessions (nothing to strip).

This is the **reverse-direction counterpart** to #137: that PR strips signed Opus thinking on the way *out* to non-Anthropic providers (`ComposedHandler`); this strips unsigned provider thinking on the way *back* to Anthropic (`NativeHandler`).

## Files

- `packages/cli/src/handlers/shared/thinking-signature.ts` — new pure helper (dependency-free, easy to unit-test).
- `packages/cli/src/handlers/native-handler.ts` — import + 1 call before the fetch, with a `log()` line when blocks are stripped.
- `packages/cli/src/handlers/shared/thinking-signature.test.ts` — regression tests.

## Testing

```
bun test packages/cli/src/handlers/shared/thinking-signature.test.ts
# 8 pass, 0 fail
bun run build   # clean
```

Tests cover: empty-signature strip, missing-signature strip, signed-thinking preservation, `redacted_thinking`/`text`/`tool_use` preservation, multi-turn counting, user-message immunity, malformed input, and the exact incident shape (`messages[].content[0]` unsigned thinking).

### DIFF
--- a/packages/cli/src/handlers/native-handler.ts
+++ b/packages/cli/src/handlers/native-handler.ts
@@ -2,6 +2,7 @@ import type { Context } from "hono";
 import type { ModelHandler } from "./types.js";
 import { log, maskCredential } from "../logger.js";
 import { wrapAnthropicError } from "./shared/anthropic-error.js";
+import { stripUnsignedThinkingBlocks } from "./shared/thinking-signature.js";
 import {
   fetchMultiModelAdvice,
   findPendingAdvisorToolResults,
@@ -127,6 +128,19 @@ export class NativeHandler implements ModelHandler {
           body: trimForLog(payload),
         });
       }
+    }
+
+    // Strip thinking blocks that arrived without a valid Anthropic signature.
+    // They originate from non-Anthropic providers (GLM/Kimi/DeepSeek reasoning
+    // surfaced as type:"thinking" with signature:"") and poison mixed-provider
+    // sessions: the Anthropic API rejects them with
+    // "messages.N.content.M: Invalid signature in thinking block" once a turn
+    // routes here. Genuine signed Anthropic thinking is preserved.
+    const strippedThinking = stripUnsignedThinkingBlocks(payload.messages);
+    if (strippedThinking > 0) {
+      log(
+        `[Native] Stripped ${strippedThinking} unsigned thinking block(s) (non-Anthropic origin) from message history for ${target}`
+      );
     }
 
     log("\n=== [NATIVE] Claude Code → Anthropic API Request ===");
--- a/packages/cli/src/handlers/shared/thinking-signature.test.ts
+++ b/packages/cli/src/handlers/shared/thinking-signature.test.ts
@@ -1,0 +1,122 @@
+/**
+ * Regression tests for unsigned thinking-block stripping on the native path.
+ *
+ * Anchored to the real incident: a mixed-provider Claude Code session
+ * accumulated GLM reasoning surfaced as `{ type: "thinking", signature: "" }`
+ * blocks, then routed a turn to a native Anthropic model. The Anthropic API
+ * rejected the request with
+ * `messages.19.content.0: Invalid signature in thinking block` (400).
+ *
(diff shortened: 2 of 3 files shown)


jsboige:
Friendly ping — this PR has been OPEN since 2026-06-02 with no maintainer review. The patch is self-contained (3 files, 182 lines, tests included) and addresses a real production bug (HTTP 400 in mixed-provider sessions when routing back to native Anthropic).

Recent context from the jsboige fork: we merged this fix into our fork as part of the sidecar consolidation (PR #2 on jsboige/claudish, MERGED 2026-06-08). It has been running in production across our cluster (6 machines) for 2 months without any reported incident.

If the patch needs adjustment to fit upstream conventions (e.g., the transform.ts dead-code resurrection pattern, or the three-tier stream-format lookup), happy to rebase. Could you take a look when you get a chance?

erudenko:
Picked this up as #182, credited to you. Closing here since I moved it into its own module.

Your diagnosis was right and the incident anchor in your tests is what made it easy to verify. Two changes worth explaining.

The strip runs at the very top of `handle()`, before the advisor block, so the advisor's request-body dump shows what actually goes on the wire rather than what arrived. Small thing, but that dump is the evidence trail for a different investigation and it shouldn't lie.

The bigger one: I spent a while convinced your fix was under-inclusive and I was wrong. My reasoning was that MiniMax and Kimi speak Anthropic's format, so `anthropic-sse` is a byte passthrough and their `signature_delta` frames reach the client with a real 64-hex signature that's non-empty (so it survives an unsigned-only filter) and still not Anthropic's (so it still 400s). I wrote 20 lines of module comment documenting that as a known gap.

Then I replayed the MiniMax fixture through the real adapter composition:

```
adapter: MiniMaxModelDialect   shouldFilterThinking: true
thinking blocks in output: 0
signature frames in output: 0
```

Thinking is stripped at emit on the Anthropic wire, for every dialect there. My probe had been built without an adapter, so the filter was off and thinking survived — I'd measured a composition that doesn't exist in production. Same result for kimi-k2.5, glm-5.2, deepseek-v4.

So a foreign signature never reaches the client, is never stored, and can never come back. Absent-signature isn't part of the problem, on the paths that reach this function it IS the problem, and the scope you chose was right the first time.

What did survive from that detour is worth having: the module now records that `shouldFilterThinking()` and this filter are coupled through the wire format, so narrowing the former silently breaks the latter.

Thanks for this. A 400 that only appears after you've switched providers mid-session is genuinely hard to attribute.

PR #142, opened 2026-06-12
fix: stream robustness, web search interception, non-streaming /compact, empty-response false-positive

jsboige:
## Summary

Consolidated fork improvements covering stream parsing, web search interception, and proxy robustness. All changes are backwards-compatible and gated behind opt-in env vars where appropriate.

### Stream Robustness (5 commits)
- **fix(stream): guarantee finalize() terminates stream** — try/catch/finally pattern in all stream parsers prevents hung streams when finalize() body throws
- **fix(stream): graceful handling of socket close / fetch errors mid-stream** — injects actionable error text instead of raw SSE error events
- **fix(stream): fix out-of-order content block lifecycle** — closes reasoning/text blocks before tool blocks to prevent "Content block not found" client errors
- **fix(stream): classify empty-response cause** — distinguishes transient vs overflow vs content-filter instead of blanket "compact" push
- **fix(stream): stop appending empty-response error after valid text content** — snapshot `textStarted` before clearing in finalize() (cluster-wide regression fix)

### Non-Streaming Support
- **fix(handler): serve non-streaming clients (stream:false) as JSON message** — `collectAnthropicSseToMessage()` buffers SSE into a single Anthropic message. Unblocks `/compact` on all proxied models (was working only on Opus/native before).

### Web Search Interception
- **feat(web-tools): route web search/fetch via MCP SearXNG** — intercepts provider web-search tool calls and GLM `<searchWeb>` tags, executes via SearXNG or MCP endpoint
- **fix(web-tools): intercept sub-agents + enhance htmlToText** — sub-agent WebSearch/WebFetch requests intercepted at proxy level; remaps to client tool_use when WebSearch is declared
- **fix(stream): WebFetch socket closure + empty response + tool_result role mismatch** — upstream fixes from earlier sessions

### Ops
- **chore: track operational utility scripts** — watchdog (hang detection + proactive restart) and capture compression scripts

### Regression Tests
- `format-translation.test.ts`: 77 tests (4 new: transient/overflow/content-filter classification + valid-text-not-empty)
- `collect-sse-message.test.ts`: 7 tests (text, tool_use, mixed blocks, empty body, malformed SSE, unparseable JSON)

## Files Changed
- `packages/cli/src/handlers/shared/stream-parsers/openai-sse.ts` — finalize() robustness, hasContent snapshot, error classification
- `packages/cli/src/handlers/shared/stream-parsers/anthropic-sse.ts` — out-of-order block fix, duplicate stop prevention
- `packages/cli/src/handlers/shared/stream-parsers/gemini-sse.ts` — finalize() try/catch/finally hang guard
- `packages/cli/src/handlers/shared/stream-parsers/ollama-jsonl.ts` — finalize() try/catch/finally hang guard
- `packages/cli/src/handlers/shared/collect-sse-message.ts` — new: SSE→message collector for non-streaming clients
- `packages/cli/src/handlers/shared/collect-sse-message.test.ts` — new: 7 regression tests
- `packages/cli/src/handlers/shared/mcp-searxng-client.ts` — new: MCP streamable-http JSON-RPC client
- `packages/cli/src/handlers/shared/web-search-executor.ts` — new: SearXNG search/fetch fallback chains
- `packages/cli/src/handlers/composed-handler.ts` — non-streaming JSON response path
- `packages/cli/src/proxy-server.ts` — sub-agent web tool interception
- `packages/cli/src/format-translation.test.ts` — 4 new regression tests

## Testing
- 84 tests pass (77 format-translation + 7 collect-sse-message)
- Live-tested on production proxy (GLM-5.1, qwen3.6, Opus, Sonnet)
- Verified: `/compact` works on all proxied models, no spurious "empty response" on valid content

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+node_modules
+dist
+.git
+logs
+*.log
+.claude
+.claudish
+ai-docs
+test-fixtures
+*.md
+!packages/cli/README.md
+.startup-shortcut-created
+Dockerfile
+docker-compose.yml
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,12 @@ validation/
 .mnemex/
 gtd/sessions/
 .agents/
+
+# Diagnostic capture (temporary, activated via CLAUDISH_CAPTURE_DIR env)
+capture/
+
+# Session artifacts (traces, deploy notes, monitoring scripts)
+trace-*.json
+kv-cache-traces.md
+monitor-traffic.sh
+deploy-notes-*.md
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -249,6 +249,24 @@ Claude Code v2.1.153+ injects `role: "system"` messages inline (e.g. system-remi
 
 Set `CLAUDISH_CAPTURE_DIR` env var to enable full request body capture for offline reproduction of hangs or malformed responses. Disabled by default (no-op when unset). Files written as JSON with metadata (timestamp, source IP, model, PID).
 
+### Opus Leak Diagnostics
+
+When VS Code displays "sonnet" or "glm" but the proxy shows `claude-opus-4-8` requests from that machine, the cause is almost always **sub-agents**, not the main session. The Agent tool spawns sub-agents that default to Opus.
+
+**Quick check:**
+```bash
+# Are there Opus requests from a specific machine?
+docker logs claudish-proxy --since 1h 2>&1 | grep "machine=HOST.*claude-opus" | head -5
+
+# Is it sub-agents? Check billing header in captures:
+docker exec -i claudish-proxy sh -c "head -c 500 /captures/req-1-NNNN-*.json"
+# Look for: cc_is_subagent=true
+```
+
+**Root cause:** Claude Code's Agent tool selects the "best available model" for sub-agents. On Anthropic direct, that's Opus. This is client-side behavior — not fixable in the proxy.
+
+**Fix:** Add a global rule in `~/.claude/rules/` instructing the model to always specify `model: "sonnet"` (or equivalent) when spawning sub-agents, and reserve Opus for genuinely complex tasks.
+
 ## Debug Logging
 
 Debug logging is behind the `--debug` flag and outputs to `logs/` directory. It's disabled by default.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -215,28 +215,44 @@ Located in `handlers/shared/stream-parsers/`:
 
 ## Web Search Interception (v7.1+)
 
-When providers emit web search tool calls (WebSearch/WebFetch) or GLM's `<searchWeb>` tags, claudish intercepts them and executes via a local **SearXNG** instance instead of forwarding to the provider (which would fail for non-Anthropic providers).
+When providers emit web search tool calls (`web_search`, `brave_web_search`, `tavily_search`) or GLM's `<searchWeb>` tags, claudish intercepts them instead of forwarding to the provider (which would fail for non-Anthropic providers).
+
+**HARD constraint**: a failed search/fetch must NEVER stop the agent. Every path degrades gracefully to well-formed text — no throws, no hung streams.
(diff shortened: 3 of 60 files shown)


erudenko:
This is a fork's worth of work squashed into one PR, and I want to handle it properly rather than just closing it, because there's real value buried in here.

Let me start with the thing I have to raise.

`web-search-executor.ts` has `const SEARXNG_URL = process.env.SEARXNG_URL || "http://search.myia.io"`. And `interceptWebTools` gets called inside `POST /v1/messages` with no env gate, before `getHandlerForRequest`. So on a default install, with nobody configuring anything, user search queries would go to that host over plain HTTP. `docker-compose.yml` hardcodes the same host.

I'm assuming that's a fork default that got carried along rather than anything intended. But it means I can't merge this branch in any form, and I'd rather say that plainly than hint at it.

Two others in the same category. `interceptWebTools` matches on English prompt prose, `/^Perform a web search for the query:\s*(.+)$/s`, so it hijacks any single-message request that happens to match, on any model, and breaks the moment Claude Code rewords anything. And composed-handler has `this.bareModelName === "qwen3.6-35b-a3b"` with a 400-word hardcoded system prompt attached to it. We have a rule against pinned model ids for exactly that reason.

Now the good part, because it's genuinely good.

You found a live bug nobody else has. `finalize()` isn't wrapped in try/catch/finally. `state.finalized = true` gets set first, and the `controller.close()` and `clearInterval` sit at the end of the body instead of in a `finally`. So a throw mid-finalize means the outer catch re-calls finalize, it early-returns as already-finalized, and we leak the ping interval and hang the request on an HTTP 200. Same shape at `openai-sse.ts:203`, `gemini-sse.ts:99` and `ollama-jsonl.ts:68`. That's roughly 30 lines to fix and it's the single most valuable finding across all 30 open PRs.

Second, `collect-sse-message.ts` and the `stream: false` branch. `ComposedHandler.handle` always returns SSE, there's no non-streaming path at all, which means `/compact` is broken on every proxied model. I hadn't realised that. Your seven tests for it are good ones.

Third, `normalizeGlmSlug`. Ten lines, anchored regex, tested, no blast radius outside GLM. Taking it.

So I'm closing this branch and landing those three separately, with credit to you on each. The anthropic-sse in-stream 429 gap is real too and I'm opening an issue for it, but implemented as a proper `sniffAnthropicStreamHead` next to the two we already have, on our 3s/15s/30s schedule rather than the six-retry five-minute loop. Ten minutes of silent stalling is worse for a user than a clear error.

One ask for next time, and I say this knowing you've sent ten PRs and most of them found something real. Send them separately. Four of the thirteen changes in here are things I want today, and they've been sitting behind nine I can't take since June. Not reviewing sooner is on me. But a small PR would have been merged months ago.

erudenko:
Actually closing now. My 9 August comment said I was closing it and then the close never fired, so this has been open under a comment saying it wasn't. Sorry for the mixed signal.

Since then I've shipped two slices of it, both credited to you:

**#173** — the non-streaming `/compact` fix. `collectSseMessage` folds the translated SSE back into a single Messages JSON when the client sent `stream: false`. Verified live on five models across all four wire formats. Two things I changed from your version: data lines match `data:` with the space OPTIONAL (the spec makes it optional and providers differ, and requiring it drops every line rather than degrading), and there's a stall guard, because buffering removes the user's ability to see a wedged turn and cancel it.

**#172** — the anthropic-sse never-hang, which was your #143.

Still outstanding from this PR and worth its own thread: the SearXNG default. `web-search-executor.ts` has `SEARXNG_URL = process.env.SEARXNG_URL || "http://search.myia.io"`, and `interceptWebTools` runs inside `POST /v1/messages` with no env gate, before `getHandlerForRequest`. On a default install that sends user search queries to that host over plain HTTP. I don't think that was the intent, but it's the reason the web-search slice can't come in as-is.

The rest I've reviewed individually: #123 and #124 are fixed on main, #125 #127 #128 #133 have specific feedback on them and are open.

PR #143, opened 2026-06-13
fix(anthropic-sse): finalize stream on mid-stream socket close (never-hang)

jsboige:
## Problem

Anthropic-compatible providers that speak native Anthropic SSE (Z.AI, MiniMax, Kimi) intermittently **close the TCP socket mid-stream** under load / burst limits. When that happens, `reader.read()` in the passthrough loop rejects, the rejection escapes to the outer `catch (e)` block, which does a **bare `controller.close()` with no terminal `message_stop` event**.

The Claude Code client then reports:

```
API Error: The socket connection was closed unexpectedly. For more information, pass verbose: true in the second argument to fetch()
```

…and **freezes the agent turn** — a hung agent, the worst proxy failure mode.

## Root cause

In [`anthropic-sse.ts`](https://github.com/MadAppGang/claudish/blob/main/packages/cli/src/handlers/shared/stream-parsers/anthropic-sse.ts), the read loop's outer catch can only raw-close the controller. There was no path to emit the terminal `message_stop` the client requires to end a turn, because a finalization helper was never in scope.

## Fix

Two changes, both additive:

1. **`finalizeWithError(errMsg, path)` helper** — declared inside the outer `try`, above the read loop. Emits a valid terminal Claude message so the client ends the turn cleanly:
   - **no content reached the client yet** → synthetic `message_start` + a content block carrying a short notice, so the closing events are well-formed
   - **content already streaming** → `message_delta(end_turn)` + `message_stop` (the client tolerates a missing `content_block_stop` on early termination far better than a missing `message_stop`, which hangs the turn)

2. **Inner `try/catch` around the read loop** — a mid-stream reader rejection is now caught *where `finalizeWithError` is in scope* and routed through it, instead of escaping to the bare outer close. The outer catch is preserved as a last-resort safety net.

Whatever streamed before the cutoff is preserved — no content loss.

## Why this matters

A frozen agent is far worse than a clean error. With providers that drop connections under load (observed in production with Z.AI serving GLM models), this turns a transient network blip into a permanently stuck session that the user must manually reset. The fix degrades it to a clean end-of-turn.

## Test

Regression test added to `format-translation.test.ts`:

> **never-hang: upstream socket close mid-stream still emits terminal message_stop**

Replays an upstream body that emits `message_start` + a text delta, then **errors** (simulating socket close). Asserts:
- the partial content survives (not lost to the cutoff)
- a terminal `message_stop` is present
- a synthesized `stop_reason: end_turn`

**Fails without the fix** (no `message_stop` reaches the client), **passes with it**.

## Scope

- 2 files, +148 lines, **additions only** — no behavior change on the happy path or the normal end-of-stream path
- Touched file: `packages/cli/src/handlers/shared/stream-parsers/anthropic-sse.ts`
- No dependencies on other PRs

### DIFF
--- a/packages/cli/src/format-translation.test.ts
+++ b/packages/cli/src/format-translation.test.ts
@@ -259,6 +259,66 @@ describe("Anthropic SSE Passthrough (createAnthropicPassthroughStream)", () => {
     expect(tokenInput).toBe(50);
     expect(tokenOutput).toBe(5);
   });
+
+  // never-hang: Z.AI and similar providers close the TCP socket mid-stream
+  // under load. Before the fix, the reader.read() rejection escaped to a catch
+  // that did a bare controller.close() with NO terminal message_stop, so the
+  // client saw "socket connection was closed unexpectedly" and froze the turn.
+  test("never-hang: upstream socket close mid-stream still emits terminal message_stop", async () => {
+    const createAnthropicPassthroughStream = await getParser();
+    const encoder = new TextEncoder();
+    // Fixture: a few valid events, then the upstream body ERRORS (socket close).
+    const upstream = new ReadableStream({
+      start(controller) {
+        controller.enqueue(
+          encoder.encode(
+            "event: message_start\n" +
+              `data: {"type":"message_start","message":{"id":"msg_test","type":"message","role":"assistant","model":"glm-5.1","content":[],"stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":10,"output_tokens":0}}}\n\n`
+          )
+        );
+        controller.enqueue(
+          encoder.encode(
+            "event: content_block_start\n" +
+              `data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}}\n\n`
+          )
+        );
+        controller.enqueue(
+          encoder.encode(
+            "event: content_block_delta\n" +
+              `data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Partial answer before cutoff"}}\n\n`
+          )
+        );
+        // Simulate the upstream TCP socket dying mid-stream — deferred so the
+        // enqueued chunks are read first (a real close arrives after a network
+        // round-trip, not synchronously in the same tick as the writes).
+        setTimeout(
+          () => controller.error(new Error("The socket connection was closed unexpectedly.")),
+          5
+        );
+      },
+    });
+    const fixture = new Response(upstream, {
+      status: 200,
+      headers: { "Content-Type": "text/event-stream" },
+    });
+    const ctx = createMockContext();
+
+    const response = createAnthropicPassthroughStream(ctx, fixture, {
+      modelName: "glm-5.1",
+    });
+
+    const events = await parseClaudeSseStream(response);
+
+    // Partial content must survive (not lost to the cutoff).
+    expect(extractText(events)).toContain("Partial answer before cutoff");
+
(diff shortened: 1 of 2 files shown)


erudenko:
This one found a real bug and it's still live on main. The outer catch at `anthropic-sse.ts:472` is a bare `controller.close()` with no terminal event, and nothing in `composed-handler.ts` synthesizes a `message_stop` for this path. So when Z.AI or MiniMax drop the socket mid-answer, Claude Code just sits there. You're right about the failure.

I can't merge the diff as it stands though, and the reasons are worth writing down.

`finalizeWithError` returns straight after finalizing, which skips main's `onTurnEnd()` and `onTokenUpdate()` at `anthropic-sse.ts:458`. Neither existed when you wrote this. So Layer 4's `finishTurn()` never fires and the partial usage never reaches the token file. We'd be trading a hang for a quieter leak.

The synthetic `message_start` drops `${opts.modelName}` straight into a hand-built JSON string. Everything else in that file goes through `JSON.stringify`. Custom endpoints allow arbitrary model names, so one quote or backslash and we emit JSON nobody can parse.

`stop_reason` is hardcoded to `end_turn` even when `stopReason` already holds the real value at `anthropic-sse.ts:391`. A turn that lost only its `message_stop` would have its tool call thrown away by the client.

And here's the thing that actually decided it. You, #133 and #127 each wrote your own version of "synthesize a terminal message_stop when the stream ends badly", three months apart, none of you talking to each other, and all three hardcode `end_turn`. Three people patching the same spot means we're missing a helper, not that we need three patches. Merging all three would put three competing finalizers on one stream.

So I'm going to build one `finalizeStream(controller, {stopReason, outputTokens, sawMessageStart, openBlockIndex})` and call it from every error path, falling through to `onTurnEnd` instead of returning past it. Your regression test goes in as-is, the deferred `controller.error` one that asserts partial text survives and `message_stop` is present. That test is the half I'd have got wrong on my own.

Credit goes on the commit. Thanks for this, it was costing people hung sessions.

erudenko:
Closing this because I've reimplemented it as #172, credited to you. The branch here had gone conflicting and I'd changed enough that a rebase would have been rewriting it.

Your diagnosis was exactly right: the catch closed the controller with no terminal event, so a socket that dropped mid-answer handed Claude Code a stream that stopped but never ended, and the turn hung.

Three things I did differently, all learned from your patch rather than instead of it.

Frames are built with `JSON.stringify` rather than string interpolation. Custom endpoints allow arbitrary model names, and one quote in a hand-built literal produces JSON the client can't parse, which turns a recoverable truncation into a hard failure. There's a test with a quote and a backslash in `modelName`.

`stop_reason` is whatever upstream actually reported, falling back to a derived value rather than a hardcoded `end_turn`. A turn that lost only its terminal frames may still have delivered a complete `tool_use` block, and reporting `end_turn` there makes the client discard the tool call.

And `onTurnEnd` / `onTokenUpdate` still run on that path, so Layer 4 and the token file see a completed turn instead of nothing. Returning early past them would have traded a hang for a quieter leak.

One thing your patch made me find: lifecycle has to be tracked at all THREE sites that emit a data frame, not just `enqueueData`. The thinking filter's renumbering path enqueues directly and passes a RENUMBERED index, so a synthetic `content_block_stop` built from the frame's own index would name a block the client never opened. That's the bug #127 exists to fix, and I'd have shipped it while fixing its neighbour.

Thanks for this one. A hang is the worst failure mode we have because it looks like the model thinking.

PR #145, opened 2026-06-17
fix(vertex): EU/US REP + global endpoints, and fix v@ routing fall-through to OpenRouter

nickoloss:
## Problem

The Vertex AI OAuth endpoint builder hardcodes a single host shape — `https://<location>-aiplatform.googleapis.com` — derived from `VERTEX_LOCATION`. That cannot address two endpoint families Vertex actually uses:

1. **Data-residency multi-region (REP)** — `https://aiplatform.<loc>.rep.googleapis.com` (e.g. `aiplatform.eu.rep.googleapis.com`, `location=eu`).
2. **Global** — `https://aiplatform.googleapis.com` with `location=global` (note: **no** `<location>-` prefix).

The practical impact: **Gemini models published only on those endpoints can't be used at all.** `gemini-3.5-flash` is currently served **only** on the EU REP endpoint and the global endpoint — never on the per-region `europe-west4-aiplatform.googleapis.com` host the code builds. So EU users have no way to run 3.5 Flash with data residency.

## Evidence (live `:countTokens` probes against a real project)

| Endpoint (host / `location`) | `gemini-3.5-flash` | `gemini-2.5-pro` | builder host before this PR |
|---|:---:|:---:|:---:|
| `aiplatform.eu.rep.googleapis.com` / `eu` | ✅ 200 | ❌ 404 | ❌ unreachable |
| `aiplatform.googleapis.com` / `global` | ✅ 200 | ✅ 200 | ❌ builds `global-aiplatform…` → 404 |
| `europe-west4-aiplatform.googleapis.com` / `europe-west4` | ❌ 404 | ✅ 200 | ✅ |
| `eu-aiplatform.googleapis.com` / `eu` (current output for `VERTEX_LOCATION=eu`) | ❌ 404 | ❌ 404 | ✅ but dead host |

So `VERTEX_LOCATION=eu` today produces `eu-aiplatform.googleapis.com`, which is not a valid Gemini host; the real EU multi-region host is `aiplatform.eu.rep.googleapis.com`.

## Change

Add `vertexApiHost(location)` and use it in `buildVertexOAuthEndpoint` (and the smoke base-URL builder):

```ts
export function vertexApiHost(location: string): string {
  if (location === "global") return "aiplatform.googleapis.com";
  if (location === "eu" || location === "us") {
    return `aiplatform.${location}.rep.googleapis.com`;
  }
  return `${location}-aiplatform.googleapis.com`; // unchanged for normal regions
}
```

The `locations/<location>` path segment is untouched in every case. Existing single-region behavior (`europe-west4`, `us-central1`, …) is byte-for-byte unchanged.

Usage after this PR:
```bash
export VERTEX_PROJECT=my-project
export VERTEX_LOCATION=eu          # data-residency multi-region
claudish --model v@gemini-3.5-flash "hi"
```

## Tests / verification

- Added `packages/cli/src/auth/vertex-auth.test.ts` covering `vertexApiHost` and `buildVertexOAuthEndpoint` for `eu` (REP), `global`, and a normal region, plus the `?alt=sse` streaming path.
- The exact URLs the helper generates for `eu`, `global`, and `europe-west4` were each confirmed to return HTTP 200 from Vertex via `:countTokens`.
- I could not run `bun test` in my environment (bun not installed here) — please confirm CI is green.

## Docs

- `docs/settings-reference.md`: documented `eu`/`us`/`global` behavior for `VERTEX_LOCATION`.
- `CHANGELOG.md`: `[Unreleased]` entry.

Refs: [Vertex AI — locations/REP endpoints](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/locations), [data residency](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/data-residency).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to [Claudish](https://github.com/MadAppGang/claudish).
 
+## [Unreleased]
+
+### Bug Fixes
+
+- **vertex**: fix `v@`/`vertex@` models silently falling through to the OpenRouter default (HTTP 401). Vertex has a static `baseUrl: ""` (its endpoint is built per-region in the vertex transport), so the empty-baseUrl filter in `getRemoteProviders()` dropped it from the registry and `resolveRemoteProvider()` returned null. Keep Vertex regardless of its empty static baseUrl.
+
+### New Features
+
+- **vertex**: support EU/US data-residency multi-region (`aiplatform.<loc>.rep.googleapis.com`) and `global` endpoints via `VERTEX_LOCATION=eu|us|global`. Enables Gemini models published only on the REP endpoint, such as `gemini-3.5-flash` with EU data residency.
+
 ## [7.5.0] - 2026-06-10
 
 ### Documentation
--- a/docs/settings-reference.md
+++ b/docs/settings-reference.md
@@ -161,7 +161,7 @@ Claudish automatically loads `.env` from the current working directory at startu
 | `POE_API_KEY` | Poe (`poe@`) | | https://poe.com/ |
 | `VERTEX_API_KEY` | Vertex AI Express mode (`v@`, `vertex@`) | | https://console.cloud.google.com/vertex-ai |
 | `VERTEX_PROJECT` | Vertex AI OAuth mode — GCP project ID | `GOOGLE_CLOUD_PROJECT` | GCP Console |
-| `VERTEX_LOCATION` | Vertex AI region | `us-central1` | |
+| `VERTEX_LOCATION` | Vertex AI region. A normal region (e.g. `europe-west4`) hits `<region>-aiplatform.googleapis.com`. `eu`/`us` hit the data-residency multi-region endpoint `aiplatform.<loc>.rep.googleapis.com` (required for models published only there, e.g. `gemini-3.5-flash`). `global` hits `aiplatform.googleapis.com` (max availability, no data-residency guarantee). | `us-central1` | |
 | `GOOGLE_APPLICATION_CREDENTIALS` | Path to GCP service account JSON file (Vertex OAuth) | | GCP Console |
 | `GOOGLE_CLOUD_PROJECT` | GCP project ID (also used by Gemini Code Assist OAuth) | `GOOGLE_CLOUD_PROJECT_ID` | |
 
--- a/packages/cli/.gitignore
+++ b/packages/cli/.gitignore
@@ -1,1 +1,5 @@
 .claudish-team-*
+
+# Build artifacts
+claudish
+*.bun-build
--- a/packages/cli/scripts/smoke/providers.ts
+++ b/packages/cli/scripts/smoke/providers.ts
@@ -8,6 +8,7 @@
 
 import type { RemoteProvider } from "../../src/handlers/shared/remote-provider-types.js";
 import { getRegisteredRemoteProviders } from "../../src/providers/remote-provider-registry.js";
+import { vertexApiHost } from "../../src/auth/vertex-auth.js";
 import type { SmokeProviderConfig, WireFormat } from "./types.js";
 
 // Providers to skip in v1 smoke tests
@@ -142,7 +143,7 @@ function getApiPath(provider: RemoteProvider): string {
 function getBaseUrl(provider: RemoteProvider): string {
   if (provider.name === "vertex") {
     const location = process.env.VERTEX_LOCATION || "us-central1";
-    return `https://${location}-aiplatform.googleapis.com`;
+    return `https://${vertexApiHost(location)}`;
   }
   return provider.baseUrl;
 }
--- a/packages/cli/src/auth/vertex-auth.test.ts
+++ b/packages/cli/src/auth/vertex-auth.test.ts
@@ -1,0 +1,76 @@
+import { describe, it, expect } from "bun:test";
+import { vertexApiHost, buildVertexOAuthEndpoint } from "./vertex-auth.js";
+
+describe("vertexApiHost", () => {
+  it("uses the classic single-region host for normal regions", () => {
+    expect(vertexApiHost("europe-west4")).toBe(
+      "europe-west4-aiplatform.googleapis.com"
+    );
+    expect(vertexApiHost("us-central1")).toBe(
+      "us-central1-aiplatform.googleapis.com"
+    );
+  });
+
+  it("uses the bare global host for location=global", () => {
+    expect(vertexApiHost("global")).toBe("aiplatform.googleapis.com");
+  });
+
+  it("uses the data-residency REP host for eu/us multi-region", () => {
+    expect(vertexApiHost("eu")).toBe("aiplatform.eu.rep.googleapis.com");
+    expect(vertexApiHost("us")).toBe("aiplatform.us.rep.googleapis.com");
+  });
+});
+
+describe("buildVertexOAuthEndpoint (google)", () => {
+  const cfg = { projectId: "p", location: "eu" };
+
+  it("targets the EU REP host while keeping the locations/<loc> path", () => {
+    expect(
+      buildVertexOAuthEndpoint(cfg, "google", "gemini-3.5-flash", false)
+    ).toBe(
+      "https://aiplatform.eu.rep.googleapis.com/v1/" +
+        "projects/p/locations/eu/publishers/google/models/" +
+        "gemini-3.5-flash:generateContent"
+    );
+  });
+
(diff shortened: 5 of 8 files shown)


nickoloss:
Update: ran the suite locally on bun 1.3.14 — the new `vertex-auth.test.ts` passes (7 pass / 0 fail), and a `bun build --compile` binary builds cleanly. The generated hosts for `eu` (`aiplatform.eu.rep.googleapis.com`), `global` (`aiplatform.googleapis.com`), and `europe-west4` were each verified to return HTTP 200 from Vertex `:countTokens`.

nickoloss:
Expanded this PR with a **second, more fundamental Vertex bug** found while testing the endpoint fix end-to-end:

**`v@`/`vertex@` models silently fall through to the OpenRouter default → HTTP 401.** Vertex's provider definition has a static `baseUrl: ""` (its endpoint is built per-region in the vertex transport). The empty-baseUrl filter in `getRemoteProviders()` (added for LiteLLM) therefore drops Vertex from the registry, so `resolveRemoteProvider("v@...")` returns `null`, `resolveModelProvider` categorizes it `"unknown"`, and the proxy never builds the Vertex handler — every Vertex request defaults to OpenRouter. Repro:

```
$ VERTEX_PROJECT=… claudish --model v@gemini-3.5-flash "hi"
[claudish] Error [OpenRouter]: HTTP 401. Check API key / OAuth credentials.
```

Fix: keep providers whose transport self-constructs the URL (`transport === "vertex"`).

**Verified end-to-end** on a compiled binary (bun 1.3.14), both fixes together:
```
$ VERTEX_LOCATION=eu VERTEX_PROJECT=… claudish --model v@gemini-3.5-flash "Reply with exactly: VERTEX_OK"
VERTEX_OK
```
That round-trips Gemini 3.5 Flash through the EU data-residency REP endpoint. Tests: 10 pass / 0 fail across `vertex-auth.test.ts` and `remote-provider-registry.test.ts`.

erudenko:
This is the one I want to land first out of everything currently open.

I checked the endpoint claim before doing anything else. `aiplatform.eu.rep.googleapis.com` answers a real generateContent call with a 401, and `eu-aiplatform.googleapis.com`, which is what we build today for `VERTEX_LOCATION=eu`, returns 404. So the host we've been constructing for the REP regions just doesn't exist. You're right.

The registry filter catch is the better half though. `getEffectiveBaseUrl(def) !== ""` in `remote-provider-registry.ts:47` drops vertex completely, because vertex has an empty `baseUrl` and builds its URL from project plus location at request time. I ran `resolveRemoteProvider("v@gemini-3.5-flash")` against main and got `null` back. So `v@` isn't degraded, it's dead, and it has been for a while. Nobody reported it, which probably tells you how few people got past the first error message.

One thing moved under you since June. The "silently falls through to OpenRouter" symptom is gone, `proxy-server.ts` now throws a RoutingError for any explicit non-openrouter spec that produced no handler. So today it fails loudly but with a misleading message about a missing credential, even when `VERTEX_PROJECT` is set. The fix is still needed, it's just worth rewording that bit of the description so the next person reading it isn't confused.

To land it: `CHANGELOG.md` is the only actual conflict, the four code and doc files all auto-merge clean. Can you drop the CHANGELOG hunk and the `packages/cli/.gitignore` churn, then rebase? I'll merge after that.

Separately, on `def.transport === "vertex"` in the filter. I'd rather that became a `selfConstructsUrl` flag on `ProviderDefinition`, since litellm already needed its own carve-out on that same line and two special cases in one predicate usually means the predicate is asking the wrong question. Not blocking this PR though, I'll do it after.

Thanks for chasing this one down. The live 404 versus 401 comparison in your description is what made it quick to verify.

erudenko:
Half of this is already on main and half is still genuinely missing, so I'd like the missing half as its own PR.

**Already fixed:** the `v@` fall-through. `vertex: vertexProfile` is registered in `PROVIDER_PROFILES` now, and `v@claude-sonnet-5` parses to `provider=vertex, explicit=true`. That whole class of bug (a definition in `BUILTIN_PROVIDERS` with no matching profile, silently routing to OpenRouter and billing there) is one of the nastier ones in this codebase, so thank you for catching it. It landed separately.

Global is there too. `vertex-auth.ts:266` builds `https://aiplatform.googleapis.com/v1/projects/…/locations/global/…` with no `<location>-` prefix.

**Still missing:** REP. Nothing matching `rep.googleapis` exists anywhere in `packages/cli/src`, so `aiplatform.eu.rep.googleapis.com` cannot be addressed at all. I checked the host is real:

```
GET https://aiplatform.eu.rep.googleapis.com/v1/  ->  404
```

404 rather than a DNS failure, so it resolves and serves, it just doesn't like an empty path. Anyone under an EU data-residency requirement currently can't use `v@`.

The endpoint builder is `vertex-auth.ts` around lines 251-266, and it's three template literals rather than one, so the REP shape slots in as a third branch. If you want to open a fresh PR against main with just that, I'll take it. What I'd want:

- the branch selected from the location value rather than a new env var, so `VERTEX_LOCATION=eu` picks REP and `global` picks the global host, matching how the existing two decide
- a test per shape, since all three are string construction and there's no way to catch a typo at runtime except by a 404 in production

I'm not able to verify a real REP call myself, no data-residency project on this account, so I'd be merging the URL shape on your say-so plus Google's docs. Worth saying out loud.

Sorry this sat. The routing half getting fixed independently is exactly the failure mode of a long queue.

erudenko:
Checking back on this one, and with something to apologise for.

Sixteen fork CI runs were sitting queued and unapproved in this repo, some since July. That means no test had ever run against your PR, or anyone else's from a fork. I've approved the backlog and the gate now covers fork PRs properly. That's most of the reason this sat.

Where things stand on the code: the `v@` routing half landed independently, so the only thing still missing is REP, and I confirmed again just now that nothing matching `rep.googleapis` exists anywhere in `packages/cli/src`. Anyone under an EU data-residency requirement still can't use `v@`.

The ask is unchanged and small. A fresh PR against `main` with only the REP branch in `vertex-auth.ts` (around lines 251-266, where the two existing shapes are built), selecting on the location value so `VERTEX_LOCATION=eu` picks REP the same way `global` already picks the global host, plus one test per URL shape since these are all string construction and a typo only surfaces as a 404 in production.

Leaving this open rather than closing it, because the finding is still live and I'd rather it stayed visible. If you'd rather not carry it further, say so and I'll write the REP branch myself with credit to you. Either is genuinely fine, I just don't want it sitting on you silently.

Still true that I can't verify a real REP call, no data-residency project on this account, so I'd be merging the URL shape on your word plus Google's docs. Worth restating.

erudenko:
Landed in #204. Thank you, and sorry this took as long as it did.

Your finding was right and the live 404-versus-401 comparison in your description is what made it quick to verify. I re-measured before writing anything:

```
eu-aiplatform.googleapis.com        404   what we built
aiplatform.eu.rep.googleapis.com    401   what exists
```

**I owe you a correction.** In my earlier review I said global was already handled and pointed at the line building the bare `aiplatform.googleapis.com` host. That line only serves non-Google publishers. A Gemini request with `VERTEX_LOCATION=global` went through the same regional template and 404'd exactly like `eu`:

```
global-aiplatform.googleapis.com    404
aiplatform.googleapis.com           401
```

So you had found two instances of one bug and I told you half of it was fine. I'd read the right file and the wrong branch.

Worth recording for anyone who touches this next: **DNS cannot tell you which of these hosts is real.** All four names resolve, because `*.googleapis.com` answers on a catch-all frontend. The broken ones even land on the generic `172.217.x` range while the REP hosts sit on dedicated `34.x` IPs. Only an actual request discriminates.

One deliberate omission. `aiplatform.us.rep.googleapis.com` also answers 401, so routing `us` there would have been easy. I left it on the regional host, which also answers 401. Both are real and they are different endpoints, and only the REP one carries a data-residency guarantee — but nothing observable from outside says which one someone setting `VERTEX_LOCATION=us` wants, and the current construction already reaches a live route. Changing it would have been inference rather than measurement. If you know that a `us` user needs REP, say so and I'll take it, but I'd rather it became an explicit opt-in than a guess from a location string.

Ten tests, one per URL shape, since a typo here is otherwise only visible as a 404 in production. Mutation-checked: removing the two special cases fails five of them.

You're co-author on the commit. Still true that I have no data-residency project to test against, so the REP shape merged on your word plus Google's docs, and I've said so in the PR.

Separately: sixteen fork CI runs had been sitting queued and unapproved in this repo, some since July, which is why nothing had ever been tested here. That's cleared.

PR #149, opened 2026-07-19
feat: classifier passthrough — run auto-mode permission classifier on native Anthropic

dingyifei:
## Summary

Adds an opt-in **classifier passthrough** so Claude Code's auto-mode permission classifier keeps working when the main loop runs on a non-Claude provider.

**Problem:** In `--permission-mode auto`, Claude Code sends a dedicated **permission-classifier** request (Anthropic's "security monitor"). It arrives with the *session* model name (e.g. `claude-opus-4-8`), so claudish routes it by role to whatever that tier maps to. In a hybrid setup like `--model-opus cx@gpt-5.6-sol`, that sends the classifier to **Codex/GPT/Gemini**, which can't act as Anthropic's safety classifier — it over-blocks or times out with *"safety classifier temporarily unavailable"*, and even harmless commands get denied.

**Fix:** Detect the classifier request by **content** (a marker in its `system` prompt) and reroute **only that request** to native Anthropic (`claude-sonnet-5` by default, via the existing `NativeHandler` + the user's Claude Max OAuth), bypassing role-based routing. Every other request follows normal routing, so the main loop stays on its mapped provider.

## Usage (opt-in, default off)

```bash
claudish \
  --model-opus cx@gpt-5.6-sol \
  --model-sonnet claude-sonnet-5 \
  --classifier-provider anthropic \
  -- --permission-mode auto -p 'refactor the auth module'
```

Enable via `--classifier-provider anthropic` / `--classifier-model <id>`, or `CLAUDISH_CLASSIFIER_PROVIDER=anthropic` / `CLAUDISH_CLASSIFIER_MODEL=<id>`. Default classifier model: `claude-sonnet-5`.

## How it works

- **Detection** (`classifier-passthrough.ts`): `isAutoModeClassifierRequest(body)` scans **all** `system` text blocks (the classifier's `system[0]` is the `x-anthropic-billing-header:` block; the marker is a later block), allocation-free, handling the string-vs-array `system` duality.
- **Reroute** (`proxy-server.ts`, `POST /v1/messages`): short-circuit *before* role routing → `rewriteClassifierForNative(body, model)` (force model + strip `thinking`) → `nativeHandler.handle(c, body)`. `NativeHandler` forwards the `system` array (incl. the billing block), the inbound OAuth `Authorization`, and `anthropic-beta` **verbatim** — no adapter change needed.
- **Auth** (`claude-runner.ts`): `shouldPreserveNativeAuth()` preserves the user's real Anthropic auth for the passthrough **only when resolvable creds exist** (native role mapping, `ANTHROPIC_API_KEY`, or `~/.claude/.credentials.json`); otherwise it keeps the main loop working and prints a warning instead of stranding the session at a login gate.

## Testing

- **Unit** (`classifier-passthrough.test.ts`) — detection (string / multi-block real shape / negatives), config precedence, payload rewrite.
- **HTTP integration** (`classifier-passthrough.integration.test.ts`) — drives the real proxy with a stubbed `fetch`: asserts reroute → native, model rewrite, `thinking` strip, and verbatim `system`/OAuth forwarding, plus negative/off cases.
- **Live E2E** (confirmed): main loop on `cx@gpt-5.6-sol`, classifier rerouted `claude-opus-4-8 → claude-sonnet-5` on `api.anthropic.com` (200), harmless write allowed, no "temporarily unavailable". `CLAUDISH_CLASSIFIER_DEBUG=1` dumps raw requests to `logs/classifier-capture.jsonl` for verification.

## Notes

- `count_tokens` is intentionally **not** hooked — its native branch authenticates with the construction-time key (not inbound OAuth); left alone, a classifier `count_tokens` routes by model and returns a harmless estimate.
- **Consent-model nuance:** Claude Code's classifier treats a command the user *explicitly names* in the prompt (e.g. literally asking it to run `curl … | bash`) as informed consent and allows it — visible blocks are for agent-initiated or hard-blocked actions. Documented in `CHANGES.md` and `docs/models/model-mapping.md`.
- Docs updated: `README.md`, `docs/models/model-mapping.md`, `docs/advanced/environment.md`, `docs/settings-reference.md`.
- `CHANGES.md` is a self-contained patch summary for reintegration — happy to fold it into `CHANGELOG.md` or drop it if you'd prefer.
- No version bump (left to maintainers' release flow).

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Claudish is a **BYOK AI coding assistant**:
 - ✅ **Autonomous mode** - Bypass all prompts with flags
 - ✅ **Context inheritance** - Runs in current directory with same `.claude` settings
 - ✅ **Claude Code flag passthrough** - Forward any Claude Code flag (`--agent`, `--effort`, `--permission-mode`, etc.) in any order
+- ✅ **Classifier passthrough** - In hybrid setups, keep Claude Code's auto-mode permission classifier on real Claude (Anthropic) while your main loop runs on another provider (`--classifier-provider anthropic`)
 - ✅ **Vision proxy** - Non-vision models automatically get image descriptions via Claude, so every model can "see"
 
 ## Installation
@@ -278,6 +279,8 @@ claudish [OPTIONS] <claude-args...>
 | `--model-sonnet <model>` | | Model for Sonnet role (default coding) | |
 | `--model-haiku <model>` | | Model for Haiku role (fast tasks) | |
 | `--model-subagent <model>` | | Model for sub-agents (Task tool) | |
+| `--classifier-provider <name>` | | Route the auto-mode permission classifier to native Anthropic (`anthropic`) | Off |
+| `--classifier-model <model>` | | Native Claude model for the classifier passthrough (also enables it) | `claude-sonnet-5` |
 | `--profile <name>` | `-p` | Named profile for model mapping | Default profile |
 | `--config <file>` | | Use this file as the run's ONLY config (replaces global + project) | |
 | `--op-env <id>` | | Load env vars from a 1Password Environment (highest priority) | |
@@ -395,6 +398,9 @@ Every one of these also accepts `CUSTOM_<NAME>_KEY` as an alias, and any of them
 | `CLAUDISH_DEFAULT_PROVIDER` | Default provider for bare model routing (v7.0.0+) | Auto-detected |
 | `CLAUDISH_QWEN_NO_THINK` | Disable thinking for Qwen models (`1`) | |
 | `CLAUDISH_NO_PREDEFINED_ENDPOINTS` | Turn the bundled endpoint catalog off entirely (`1`) | Catalog on |
+| `CLAUDISH_CLASSIFIER_PROVIDER` | Route auto-mode permission classifier to native Anthropic (`anthropic`) — see [Classifier Passthrough](docs/models/model-mapping.md#auto-mode-classifier-passthrough) | Off |
+| `CLAUDISH_CLASSIFIER_MODEL` | Native Claude model for the classifier passthrough (also enables it) | `claude-sonnet-5` |
+| `CLAUDISH_CLASSIFIER_DEBUG` | Dump raw requests to `logs/classifier-capture.jsonl` for classifier-detection debugging (`1`) | Off |
 
 #### Claude Code Compatibility
 
--- a/docs/advanced/environment.md
+++ b/docs/advanced/environment.md
@@ -115,6 +115,8 @@ export CLAUDISH_CLASSIFIER_MODEL=claude-sonnet-5
 
 Set to `1` to dump each request's raw model / sampling params / system prompt / headers to `logs/classifier-capture.jsonl` for troubleshooting detection. Off by default.
 
+Off by default deliberately: the captured `system` array is the **full** system prompt, which carries your CLAUDE.md and project rules, and the classifier request additionally names the command being classified. Credentials are recorded only as `<present>`, never as values. Turn it on to diagnose a drifted detection marker, then turn it back off and delete the file.
+
 ```bash
 export CLAUDISH_CLASSIFIER_DEBUG=1
 ```
--- a/docs/advanced/environment.md
+++ b/docs/advanced/environment.md
@@ -90,6 +90,37 @@ export CLAUDE_CODE_SUBAGENT_MODEL='...'
 
 ---
 
+## Classifier Passthrough
+
+Route Claude Code's auto-mode permission classifier to native Anthropic while your main loop
+runs on another provider. Full guide: [Model Mapping → Auto-Mode Classifier Passthrough](../models/model-mapping.md#auto-mode-classifier-passthrough).
+
+### `CLAUDISH_CLASSIFIER_PROVIDER`
+
+Set to `anthropic` to enable classifier passthrough (default off). CLI equivalent: `--classifier-provider anthropic`.
+
+```bash
+export CLAUDISH_CLASSIFIER_PROVIDER=anthropic
+```
+
+### `CLAUDISH_CLASSIFIER_MODEL`
+
+Native Claude model the classifier request is rewritten onto. Setting it also enables passthrough. Defaults to `claude-sonnet-5`. CLI equivalent: `--classifier-model <id>`.
+
+```bash
+export CLAUDISH_CLASSIFIER_MODEL=claude-sonnet-5
+```
+
+### `CLAUDISH_CLASSIFIER_DEBUG`
+
+Set to `1` to dump each request's raw model / sampling params / system prompt / headers to `logs/classifier-capture.jsonl` for troubleshooting detection. Off by default.
+
+```bash
+export CLAUDISH_CLASSIFIER_DEBUG=1
+```
+
+---
+
 ## Network Configuration
 
 ### `CLAUDISH_PORT`
--- a/docs/models/model-mapping.md
+++ b/docs/models/model-mapping.md
@@ -94,6 +94,79 @@ Yep. Claudish detects this is an Anthropic model ID and routes directly to Anthr
 
 ---
 
+## Auto-Mode Classifier Passthrough
+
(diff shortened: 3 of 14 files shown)


dingyifei:
tldr: This PR, generated using Claude Code, allow you to use API/Codex/etc in combination with a claude subscription so auto mode works. Technically you don't need a claude sub, but claude sonnet-5 is reliable as a classifier

erudenko:
This is a real gap and I like how you solved it. Rerouting the one request the foreign model should never see, instead of trying to supervise what it does afterwards, is the right call.

I checked main and nothing covers this. `POST /v1/messages` routes purely on `body.model`, there's no content-based short circuit anywhere in there.

Four things before I can merge.

The `await` matters. Your line is `return nativeHandler.handle(c, body)`, which matches the old code, but main changed the sibling line to `return await handler.handle(c, body)`. There's a comment above it at `proxy-server.ts:869` saying why. Without the await a rejection escapes the try/catch and ends up in Hono's error handler.

`DEFAULT_CLASSIFIER_MODEL = "claude-sonnet-5"` is a pinned model id. We have a rule against those. They go stale and we've been bitten by it more than once. Derive it from the session's native role mapping or the catalog, and keep the literal as the last resort.

The third one is subtle and would break your feature quietly. `shouldPreserveNativeAuth()` returns true when it sees `ANTHROPIC_API_KEY`. But `claude-runner.ts:1201` now deletes that env var unless `wantsAnthropicApiBilling(config)` is set. So a user who exports the key specifically to feed the classifier gets it stripped, and the classifier 401s. That guard needs widening, not just a swapped check.

Last one is small. Move `CLASSIFIER_SYSTEM_MARKER` into `behavior/harness.ts`. That's where Claude Code prompt anchors live, `PLAN_MODE_HINT` is sitting right there. Same kind of thing, same reason to keep them together.

Two bits I want left exactly as they are. The allocation-free `startsWithMarker`, because avoiding `trimStart()` on a multi-KB system block on the per-request path is the sort of thing people usually skip. And `spawnSync security find-generic-password` without `-w`, so it checks the key exists without prompting the Keychain or ever reading the token. That's careful work and I noticed.

One ask on `maybeCaptureClassifierRequest`. It's env-gated and writes to its own file, which is right. But a system array carries CLAUDE.md and project rules, so please say that plainly in the docs.

The rebase is smaller than it looks, three spots really, the rest is docs and tests. Ping me when it's up and I'll take another pass.

erudenko:
This is the most carefully documented PR in the queue and the problem is real. Nothing on main handles it, so in `--permission-mode auto` with a hybrid setup the security classifier does run on whatever foreign model the tier maps to. I want this.

Three things before it can land.

**Your detection rationale is stale, though the code survives it.** You scan all `system` blocks rather than `system[0]` because "the classifier request's `system[0]` is a separate `x-anthropic-billing-header:` block". That block no longer exists. `228db3c` strips the billing header in `transform.ts`, and the array path runs the strip BEFORE the empty-block filter, so a block that was nothing but the header is removed rather than blanked. So `system[0]` is now likely the security-monitor prompt itself.

Scanning all blocks still works, and is still the right call. But the reason in the PR describes a shape that's gone, and the next person to touch this will go looking for a header block that isn't there.

**Marker-string detection fails OPEN, on a security control.** If Anthropic rewords the classifier prompt by one character, `isAutoModeClassifierRequest` returns false, the request routes to the foreign model, and nothing anywhere says so. The user believes they have passthrough enabled and they silently don't.

I'd want the opposite bias. If passthrough is enabled and a request looks classifier-shaped but doesn't match the marker, say so loudly once per session. Shape is a weaker signal than the string but a decent tripwire: `--permission-mode auto` classifier calls have a recognisable profile (small `max_tokens`, tool-less, one user turn). It doesn't have to be a second detector, just something that notices "enabled, and yet zero classifier requests this session" and prints a line.

**And it needs a rebase.** 15 files, and `proxy-server.ts` / `claude-runner.ts` have both moved. `POST /v1/messages` in particular now has the `[RequestMeta]` log and an `await`ed `handler.handle` (the `await` is load-bearing — without it a rejection escapes the try/catch into Hono's error handler).

One thing I want to say plainly: routing a security decision to Anthropic while the work runs elsewhere is the right instinct, and the opt-in default is right too. I'd rather ship this with a loud failure mode than not ship it.

dingyifei:
Thank you for your feedback! @erudenko

I'll look into this PR once I got more time in the next week or two. From my understanding the main concerns are:

1. hard coded stuff
2. better documentation
3. failure mode
4. claude-runner.ts:1201 behavior change
5. rebase

Another big news that happened recently on auto mode billing: `The auto mode classifier uses a small number of extra tokens per tool call, and we're no longer charging Claude Code users on Pro, Max, and Team plans for that classifier overhead, effective today.`

https://claude.com/blog/auto-mode-default-in-claude-code

This feature will probably need a huge warning somewhere given Anthropic no longer count auto mode classifier in billing.

erudenko:
Checking back with an apology and one piece of news.

The apology: sixteen fork CI runs were sitting queued and unapproved in this repo, two of them yours from 19 July. So no test had ever run against this PR. I've cleared the backlog and fork PRs are covered by the gate now.

The news, which changes the shape of your point about billing. You flagged that Anthropic stopped charging Pro/Max/Team users for classifier overhead. I think that strengthens this PR rather than weakening it, and it's worth saying why so the framing in the description can be updated.

The cost argument for routing the classifier away from Anthropic is now gone. What's left is the correctness and security argument, which was always the stronger half: under `--permission-mode auto` with a hybrid setup, a permission decision is being made by whatever foreign model the tier maps to, and the user has no idea. That's true whether or not anyone is billed for it. So the PR should probably lead with that and treat the billing note as a footnote, since a reader who sees "saves tokens" as the headline will now discount the whole thing.

The four changes I asked for are unchanged, and the one I care most about is still the marker-string detection failing open. A security control that silently turns itself off when Anthropic rewords a prompt is worse than one that isn't there, because the user believes it's on. I'd take this with a loud failure mode well before I'd take it with a quiet one.

No rush from my side, and thanks for saying you'd get to it. Leaving it open.

erudenko:
Merged, rebase not squash, so both your commits are on main with your authorship.

The rebase touched six files. Five were additive and boring. The auth gate in `claude-runner.ts` wasn't, so I want to flag what I changed there.

Main grew a rule since July: in the native-Anthropic branch it now deletes a real `ANTHROPIC_API_KEY` unless you opt into metered billing, because an incidental key sitting in a `.env` silently switches the session off the claude.ai subscription and you find out on the invoice. That rule quietly assumes a subscription is there to fall back on. Your `shouldPreserveNativeAuth` widens the same branch to cover classifier passthrough, and for a pure-Codex user whose only Anthropic credential is that key, the two together would have deleted the exact credential the passthrough exists to use. Login gate, which is what your guard was written to prevent.

So the deletion is now gated on `hasNativeAnthropicMapping` rather than on the whole preserve condition. Native behaviour is unchanged. Under classifier-only passthrough the key survives, and the exposure is small since the main loop is on a foreign provider, so the only thing that key can bill is the classifier call itself.

That was a one-line condition buried in `runClaudeWithProxy`, which spawns a process and can't be tested. I pulled it out as `shouldHideIncidentalAnthropicKey` and pinned all five cases in `force-login-method.test.ts`, including the classifier-only one. `wantsAnthropicApiBilling` takes an optional env now so the injected env actually reaches it, and the new block points `setConfigFileOverride` at a temp config, otherwise anyone with `anthropicApiBilling` set for real fails the default case.

Two other things. Dropped `CHANGES.md` as you offered, it's all in the commit messages. And I added a docs line: `CLAUDISH_CLASSIFIER_DEBUG` writes the full `system` array, so CLAUDE.md and project rules land in `logs/classifier-capture.jsonl`. Worth saying next to the switch instead of leaving people to find out.

I also checked your detection anchor against 2.1.234 before merging. `You are a security monitor for autonomous AI coding agents.` is still in the binary verbatim, and `--permission-mode auto` is still there, so the marker hasn't drifted in the seven versions since you wrote this.

The keychain probe is my favourite part. `security find-generic-password` with no `-w` so nothing prompts and no secret is read, and an injectable deps seam instead of mutating `process.platform` in tests. That's the right shape and I'd have got it wrong on the first try.

PR #158, opened 2026-08-07
feat(session): end-of-session summary card + worktree-aware resume picker

erudenko:
Claude Code prints `Resume this session with: claude --resume <id>` when it exits. Claudish printed `[claudish] Done`. This adds both halves — a stats card, and a picker for the id.

## The card

Static ANSI printed after the proxy shuts down, so it lands in the scrollback intact — the same split `probe-results-printer.ts` uses, for the same reason.

```
╭─ session ──────────────────────────────────────────────────────────────────╮
│  google@gemini-3.1-pro   EST                                       12m 34s │
│ OpenRouter                                                                 │
│ context   ████████████████░░░░░░░░░░░░░░░░░░░░░░░░        39%  412.0K/1.0M │
│ tokens    [in ████████████████████████████████████▏out]  in 412.0K out 18.4K │
│ spend     [in ██████████████▏out ██████████████████████]  in $0.206 out $0.037 │
│ tools     [Read▏Edit▏Bash▏Grep▏Write]                              47 calls │
│           Read 18 · Edit 12 · Bash 9 · Grep 5 · Write 3                    │
│ cost      $0.243  estimated                                                │
│ vs Sonnet ██████████████████████████████░░░░░░░░░░        76% saved $0.765 │
│ vs Opus   ████████████████████████████████████░░░░         90% saved $2.28 │
╰────────────────────────────────────────────────────────────────────────────╯

Resume this session with:
claudish --model google@gemini-3.1-pro --resume 8c5f68cf-005a-48e3-bf25-27e046e5f879
```

**The resume line is outside the box and unstyled on purpose.** Inside it, a terminal row-select drags the `│` borders along and the pasted command will not run.

**`tokens` and `spend` are a pair.** Tokens are ~96% input, so that bar alone is a flat rectangle — the aesthetic contract's explicit failure case. Output is priced 4–5× higher, so the spend bar inverts it. Together they say: the few tokens the model wrote are where the money went.

**Savings are live, never constants.** Resolved through the catalog's `~anthropic/claude-<tier>-latest` alias rules (today Sonnet 5 at 2/10, Opus 5 at 5/25), taking the `anthropic` aggregator's rate specifically — a gateway resale rate would inflate the figure. No catalog → no savings line, never a guess. A model that cost *more* renders `over by $X` rather than a negative saving.

## The picker (bare `--resume`)

`--resume <id>` still passes through untouched; only the id-less form is intercepted. Forwarded bare, Claude Code lists the current directory's sessions only — in a worktree, the one place the wanted session is not.

Worktrees left, sessions right, preview below, fuzzy filter over both. Rows are labelled with Claude Code's own `ai-title` records, so you read "Fix Gemini answers silently deleted by reasoning regex", not a UUID. Deleted worktrees still appear (their transcripts remain resumable), marked as such.

**Scale drove the design.** Measured on one machine: 5,656 transcripts, 2.57 GB, largest single file 73.5 MB. Reading transcripts to build a list is off the table.

- Listing is directory-names + `stat` only — 1,907 sessions in ~45 ms.
- Detail hydrates lazily from a 64 KB head + 128 KB tail via positioned reads, so a 73 MB transcript costs what a 73 KB one does.

## Fixes a real pre-existing bug

`openai-sse.ts` declared `onToolCallObserved` and never called it — the parser serving GLM, Kimi, Grok, DeepSeek, Qwen, OpenRouter and LiteLLM. The behaviour layer's tool-name list was always empty on that wire.

Hooked at the single frame writer rather than the seven `content_block_start` sites, so a new tool_use path cannot forget to opt in and cannot double count. Tool counting is also decoupled from `behaviorSession`, which was gating it off entirely whenever the behaviour layer was disabled.

## Vendored viz layer (`tui/viz/`, `tui/runtime/`)

Colour maths, cell-width maths and the six widgets from the `bunjs:opentui-tui` skill, byte-identical except `tokens.ts`, which re-points the palette at claudish's existing `C` so there is one palette rather than two. Three adaptations, each forced by measurement rather than preference:

- **`chromeIntent` dropped.** It calls `RGBA.defaultForeground`, which does not exist in our pinned `@opentui/core@0.1.87` — it threw at *module import*, taking down all 25 widget tests before one ran. 0.1.x is pinned deliberately (`build:binary` uses `--compile`).
- **`bgPanel` is `C.bgAlt`, not `C.bg`.** A panel background equal to the terminal void is not a background — the test helper's own comment says "unpainted terminal comes back 0,0,0", and with a black panel it counted 12 rows of a 3-row panel.
- **A dead import removed**, caught by claudish's `noUnusedLocals`, which the skill's own tsconfig does not enable.

## Verification

- `bun test` — **1961 pass, 5 skip, 0 fail**
- `bunx tsc --noEmit` — clean
- `check-surface.ts` (the skill's core/React surface linter) — clean
- Colour screenshots read at **80×24 and 145×45**; the card rendered for typical / free / over-budget+failed states
- Picker driven by real keystrokes: Escape and Ctrl+C cancel, Enter picks, arrows pick a different session

## Not included

Unit tests for the new `session/` modules. Per project convention test-writing is delegated to Codex, which is out of credits until 2026-08-08.

The brief lives locally under `ai-docs/sessions/` (gitignored — session artifacts are deliberately not tracked). What it needs to cover, so it is not lost:

- **`ansi-viz`** — `meter()` is exactly `width` visible cells at pct 0/50/100/150/-20; `meter(NaN)` differs from `meter(0)` (a dead sensor must not look like a healthy 0%); the fill is a real gradient (>= 12 distinct fg colours at width 40); `stackedBar()` is exactly `width` columns for every mix including all-zero and over-subscribed; `visibleWidth()` counts CJK and emoji as 2 columns, which is the bug `.length` has.
- **`session-stats`** — drive via `CLAUDISH_TOKEN_FILE`; `context_window: "unknown"` yields `null`, not `NaN`; `durationMs` is never negative under clock skew; `inputCostUsd + outputCostUsd === costUsd` exactly, including on the actual-cost path where the rescale matters.
- **`session-summary`** — the resume line is last, starts with `claudish --model `, and is completely free of ANSI (the whole point); every boxed line has identical visible width; `savedUsd < 0` renders "over by" and never "saved".
- **`session-discovery`** — `slugForPath` doubled-dash case; `ACTIVE_WINDOW_MS` boundary; `sessionLabel` precedence title > firstPrompt > id.

Constraint worth repeating in any test written here: **never `mock.module()`** — Bun's registry bleeds across files and breaks sibling e2e suites.


erudenko:
@claude  review please

erudenko:
All four addressed in d746cf2. Two were real bugs that would have shipped wrong numbers — thanks, the savings one in particular was well spotted.

**1. Savings basis — confirmed and fixed.** Verified in `token-tracker.ts`: `sessionInputTokens =` (assignment, 5 sites) against `sessionOutputTokens +=` and `sessionTotalCost +=`. The baseline priced one turn of input and subtracted N turns of spend, so the sign really could invert.

Added `sessionBilledInputTokens`, accumulated at each of the five charging sites to mirror exactly what that strategy bills — the delta on delta-aware, the full context on the assignment paths, the running total on accumulate-both (assigned, not `+=`, since that path already sums input itself and would otherwise double-count), and the real volume on the free/local path, because "what this would have cost on Sonnet" is the entire point of the comparison for a free model. Published as `billed_input_tokens`; `session-stats.ts` falls back to `input_tokens` only for token files predating the field.

**2. Worktree matching — confirmed both directions, restructured.** Ownership is now decided against ground truth rather than a slug prefix: git's worktree list first (which is what makes a non-nested `git worktree add ../repo-feature` visible at all), then the `--claude-worktrees-` marker for deleted ones, then the root itself. Only the genuinely ambiguous case — a slug extending the root's with no marker — consults the transcript's recorded `cwd`, and `isUnder` is path-segment aware so `/repo-old` is not under `/repo`.

Checked the cost, since reading transcripts to classify was the thing the module was built to avoid: 17 ambiguous directories out of 225 here, **8ms** for all the probes. Discovery is 10–17ms warm, down from 44ms.

One gap I left in rather than papered over, now documented at the call site: a **deleted** worktree that lived outside the root stays invisible — git no longer lists it and its slug does not extend the root's, so nothing points at it short of the full 225-directory scan this design exists to avoid.

**3. Devin `connect-proto` — confirmed, fixed.** Worth noting why it was missed: that fifth case arrived with the rebase onto main (Devin support is in those 46 commits), so it did not exist when the sweep ran. Now uses the shared `observeToolCall`.

**4. Resume line — fixed, slightly differently than suggested.** Rather than reconstructing a routed spec, `resumeModelSpec` is now separate from the display label and is only ever the user's explicit `--model`. Absent that, `--model` is **omitted** and the line is `claudish --resume <id>`, which re-reads the same profile. A role-mapped session has no single spec that could faithfully represent it, so omitting is correct rather than merely safe — and it removes the wrong-provider risk entirely instead of narrowing it.

`bun test` 1961 pass / 5 skip / 0 fail, `tsc --noEmit` clean.

erudenko:
Ran a second, **independent** review as well (fresh reviewer, no context on the work). It found things the first pass did not. Five fixed in 27b938f; two did not survive verification and I want to be explicit about those rather than quietly skip them.

### Fixed

**The picker replaced `--resume` instead of upgrading it** (blocking, two ways):
- `claudish -p --resume … | jq` opened an alternate screen into the stdout print mode reserves for machine-readable output, then blocked forever on a keypress that could not arrive.
- Outside a git repo `getRepoContext` returns null (verified in `/tmp`), so the picker reported "no sessions" and exited 0 — **Claude Code never launched**. That was a straight regression: the flag used to be forwarded and you got Claude Code's own picker.

Now any path that cannot draw a TUI — no TTY, print mode, non-interactive, or nothing to group — forwards the bare flag. Cancelling still exits 0 rather than reopening a picker you just dismissed.

**Card rows could overflow and destroy the border.** `padVisible` padded but never clipped, so one long `mcp__server__tool_name` in the legend ran 21 columns past a 64-column card and broke the right border for every row below. Added `clipStyled` (ANSI-aware, passes SGR runs through at zero width, never splits a wide glyph) and made `padVisible` clip as well as pad, so the invariant is structural instead of per-caller. Verified: one distinct boxed-row width at 64/80/120 columns with a pathological tool name.

**The resume line could name someone else's conversation.** Newest-by-mtime is only "necessarily" this session when one session owns the directory — a second terminal, a `team` run or an MCP channel child can be newer at exit. Now prefers the id claudish actually launched with, and bounds the mtime guess to the session's own lifetime so it fails closed.

**Dead sibling worktrees could merge** — `resume-message` folded into `resume`, losing its heading. The fold now confirms real path descent via the recorded `cwd`.

**Stale tail on the live session** — `hydrateSession` used the size captured during the list pass, so a still-appending session (green dot, most likely pick) had its tail window aimed at the middle of the file. Active rows are re-stat'd first.

### Not changed — verified against, not skipped

**"A cross-worktree pick yields an id the child cannot resolve."** Flagged as the top blocking issue, on the reasoning that `claude --resume` resolves against `~/.claude/projects/<slug(cwd)>/`. The reviewer noted they could not execute it to prove it. I did, and it is **refuted**: a session recorded under `-Users-jack-mag-claudish`, resumed from a cwd slugging to `…--claude-worktrees-resume-message`, returned `OK`. `--resume <id>` resolves globally. The picker's headline capability works.

**"Per-keystroke rebuild of 610 rows; 500ms blank screen."** Measured instead: first paint **59ms**, re-renders **3–7ms** at 611 sessions. And discovery runs *before* `createCliRenderer`, so its cold cost is a normal CLI pause, not a blank alternate screen. Hand-rolled windowing would trade that for the overstrike failure the `opentui-tui` skill documents, so I left it.

`bun test` 1961 pass / 5 skip / 0 fail, `tsc --noEmit` clean, card re-screenshotted at three widths.

PR #161, opened 2026-08-07
fix(ci): check out the PR head for @claude comment-triggered reviews

erudenko:
`@claude review please` on a PR has never been able to review a PR that **adds** files. Found while trying to get a review on #158.

## What happens

For an `issue_comment` event GitHub's context points at the **default branch**, so `actions/checkout` with no `ref:` checks out `main`. The action's data fetcher then tries to read each changed file, every newly-added path is missing from that tree, and the run dies before Claude sees anything:

```
fatal: could not open 'packages/cli/src/session/session-discovery.ts'
       for reading: No such file or directory
    at fetchGitHubData (.../src/github/data/fetcher.ts:477)
```

Observed on #158 (16 new files): two runs, ~3m each, `num_turns: 1`, `total_cost_usd: 0`, `is_error: true`, no review posted — just "Claude encountered an error".

The two triggers behave differently, which is what hid this:

| Trigger | Checkout | Result |
|---|---|---|
| `pull_request` | `refs/pull/N/merge` ✅ | fine (but only fires with the trigger phrase / label) |
| `issue_comment` | `-B main refs/remotes/origin/main` ❌ | dies on any newly-added file |

A PR that only edits existing files reviews fine either way, so the bug is invisible until someone adds a file.

## The fix

One `ref:` on the checkout step, scoped to the comment-triggered case. `pull_request` events already resolve correctly and are left alone.

## Caveat worth knowing

This only takes effect **once it is on `main`** — workflows for `issue_comment` always run the workflow file from the default branch, not from the PR branch. So this PR cannot demonstrate its own fix; #158 could not be reviewed by comment until this lands.

### DIFF
--- a/.github/workflows/claude-code.yml
+++ b/.github/workflows/claude-code.yml
@@ -30,6 +30,19 @@ jobs:
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
+          # For an `issue_comment` event GitHub's context points at the DEFAULT BRANCH,
+          # not the PR — `actions/checkout` then checks out `main` and the action reviews
+          # the wrong tree. It is invisible on a PR that only edits existing files, and
+          # fatal on one that ADDS them: the action's data fetcher tries to read each
+          # changed file, every new path is absent, and the run dies with
+          # `fatal: could not open '<new file>' for reading` before Claude sees anything.
+          # Observed on #158 (16 new files): two runs, ~3m each, is_error with no review.
+          # `pull_request` events already resolve to refs/pull/N/merge, so this only
+          # needs to fill in the comment-triggered case.
+          ref: >-
+            ${{ github.event_name == 'issue_comment'
+                && format('refs/pull/{0}/head', github.event.issue.number)
+                || '' }}
 
       - name: Claude Code Action
         uses: anthropics/claude-code-action@v1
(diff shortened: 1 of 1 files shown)


erudenko:
Pushed three follow-ups from the review of this PR.

**`pull_request_review_comment` had the same bug.** It triggers the job, but the `ref:` expression only handled `issue_comment`, so an inline @claude review comment still checked out the default branch and still died on any file the PR adds. The two events carry the PR number in different places, `github.event.issue` versus `github.event.pull_request`, and review comments have no `issue` object at all.

**Added an author_association gate on both comment events.** They run in the base repo context, so unlike a fork `pull_request` they do get `ANTHROPIC_API_KEY` and a token with pull-requests/issues write. Now that checkout resolves to `refs/pull/N/head`, that pairs repository secrets with contributor-controlled content. Nothing executes the checked-out tree so it isn't RCE, but the action reads those files and prompt injection is a real path to making it act with that write token. claude-code-action does its own actor check at runtime, this just makes the boundary explicit in the workflow instead of depending on the action's internals.

**Skip the job on fork `pull_request` events.** They get no secrets so the action can only fail. #101's only failing check was exactly this, a guaranteed red X on a contribution whose code was fine.

Also worth noting for whoever merges: the earlier CI failure on this PR was a stale base, not the diff. It ran against `4e44851`, four commits behind, and the three failures (`magmux not found`, the MiniMax casing assertion, the effort-mapping assertion) are all fixed on main by `9d3e5cf` and `1325256`. Both of those files pass locally against current main.

PR #174, opened 2026-08-10
fix(openai): give codex models a Responses body, not just a Responses endpoint

erudenko:
Every `oai@*codex*` model returned 400 on its first request:

```
Unsupported parameter: 'messages'. In the Responses API, this parameter has moved to 'input'.
```

## The endpoint was already right

`OpenAIProviderTransport` has tested `modelName.toLowerCase().includes("codex")` since long before this — it picks the endpoint (`transport/openai.ts:37`) and the stream format (`:31`). So Layer 3 was already sending every codex id to `/v1/responses`.

Layer 1 disagreed. `requiresResponsesApi` matched `^gpt-5.6` only, so `openaiProfile` built an `OpenAIAPIFormat` and put a **Chat Completions body on a Responses endpoint**. Right endpoint, wrong body — and two layers had quietly drifted apart on a rule they both implement.

Widening the gate to the *same substring test the transport already uses* brings them back into agreement. That's why it's a substring and not a prefix or an id list: any narrowing re-opens the split, because the transport keeps routing ids this function would no longer claim. A test now pins that agreement.

## Evidence

Straight against `api.openai.com` with `gpt-5.3-codex`, no claudish in the picture:

| endpoint | result |
|---|---|
| `POST /v1/chat/completions` | **404** — *"This model is not supported in the v1/chat/completions endpoint. Use the v1/responses endpoint instead."* |
| `POST /v1/responses` | **200** |

So this isn't the `gpt-5.6` tools-plus-reasoning constraint wearing a second face. Codex models are flatly Responses-only, and Claude Code sends tools on essentially every request — so they were **unusable** through `oai@`, not degraded.

Through claudish, same model, A/B on the gate alone:

| gate | result |
|---|---|
| with the `codex` clause | **200** `text/event-stream` |
| without it | **400** `unsupported_parameter` |

## A trap worth knowing about

`/v1/models` is a **catalogue, not a served set**. It lists six codex ids; only `gpt-5.3-codex` answers 200 on `/v1/responses` for my account:

| id | `/v1/responses` |
|---|---|
| gpt-5.3-codex | **200** |
| gpt-5.2-codex | 404 Model not found |
| gpt-5.1-codex | 404 Model not found |
| gpt-5.1-codex-max | 404 Model not found |
| gpt-5.1-codex-mini | 404 Model not found |
| gpt-5-codex | 404 — also **deprecated** on chat/completions |

Testing against any of the others looks exactly like "the routing fix didn't work" when the model is simply not being served. `gpt-5-codex` is what I was originally going to verify against, which would have been a dead end. Same catalogue-vs-served-set trap the Antigravity and Devin providers already document — it's in the source comment so the next person doesn't spend the afternoon on it.

## Tests

16 tests on the existing hermetic `describeHandler()` seam — no keys, no network. Reverting the clause fails **7** of them.

One replaced a false positive I nearly shipped: *"gpt-5.4 and gpt-5.4-codex get different compositions"* passes **with or without** the fix, because the transport already gives the codex id a different endpoint. The assertion was satisfied for the wrong reason. It's now a layer-agreement test asserting stream format **and** endpoint together — which is what actually catches the split — plus a non-codex negative control.

Full hermetic suite: 2180 pass, 0 fail.

Still a temporary name gate, still slated for replacement by the catalog capability record once route-time capability fetch lands.

### DIFF
--- a/packages/cli/src/providers/provider-profiles.test.ts
+++ b/packages/cli/src/providers/provider-profiles.test.ts
@@ -95,3 +95,66 @@ describe("OpenCode Zen provider composition", () => {
     }
   }
 });
+
+describe("OpenAI Responses-API gate", () => {
+  const responsesComposition = {
+    transport: "openai",
+    streamFormat: "openai-responses-sse",
+    endpoint: "https://api.openai.com/v1/responses",
+  } as const;
+  const chatCompletionsComposition = {
+    transport: "openai",
+    streamFormat: "openai-sse",
+    endpoint: "https://api.openai.com/v1/chat/completions",
+  } as const;
+  const openAICompositionCases = [
+    { model: "gpt-5.3-codex", expected: responsesComposition },
+    { model: "gpt-5-codex", expected: responsesComposition },
+    { model: "gpt-5.1-codex-max", expected: responsesComposition },
+    { model: "gpt-5.1-codex-mini", expected: responsesComposition },
+    { model: "GPT-5.1-CODEX-MINI", expected: responsesComposition },
+    { model: "gpt-5.6-luna", expected: responsesComposition },
+    { model: "gpt-5.6-sol", expected: responsesComposition },
+    { model: "gpt-5.4", expected: chatCompletionsComposition },
+    { model: "gpt-5.5", expected: chatCompletionsComposition },
+    { model: "gpt-4o", expected: chatCompletionsComposition },
+    { model: "gpt-5", expected: chatCompletionsComposition },
+    { model: "gpt-5-mini", expected: chatCompletionsComposition },
+  ] as const;
+
+  for (const row of openAICompositionCases) {
+    test(`openai ${row.model} uses ${row.expected.streamFormat} at ${row.expected.endpoint}`, () => {
+      expect(describeHandler("openai", row.model)).toEqual(row.expected);
+    });
+  }
+
+  test("a codex id routes to Responses regardless of where 'codex' appears in the name", () => {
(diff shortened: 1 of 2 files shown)


erudenko:
Credit where it's due: @bassner got here first in #130, which contains the same `provider-profiles.ts` change:

```diff
-    const adapter = new OpenAIAPIFormat(ctx.modelName);
+    const isCodexModel = ctx.modelName.toLowerCase().includes("codex");
+    const adapter = isCodexModel
+      ? new CodexAPIFormat(ctx.modelName)
+      : new OpenAIAPIFormat(ctx.modelName);
```

Same substring test, same conclusion, months earlier. I found this while triaging #130 after opening this PR, not before, so this is a correction rather than an acknowledgement.

The implementations differ in one way worth recording. #130 branches on the model name at the adapter construction site; this routes through `requiresResponsesApi()`, which also swaps the endpoint and picks up the existing `gpt-5.6` case. Both land on CodexAPIFormat.

jsboigeEpita:
Closing — duplicate of #130 (bassner), as @erudenko pointed out. The Responses body slicing is already covered by the upstream implementation. Thanks for the redirect.

jsboige:
Closing — duplicate of #130 (bassner), as @erudenko pointed out. The Responses body slicing is already covered upstream. Thanks for the redirect.

jsboige:
From our side this reads as covered by the upstream provider-profiles Responses slicing (and close to bassner's #130, which you pointed at earlier). Happy to be wrong — but if it is a duplicate, closing is fine by us.

PR #180, opened 2026-08-11
feat(proxy): /v1/models discovers routable models when servedSlotIds unset

jsboige:
Reshapes #128 to the coexistence pattern from your review.

## What this does

`/v1/models` now serves two coexisting consumers from one response shape:

1. **Claude Desktop** in third-party-inference mode: when `servedSlotIds` is supplied (the `serve` command), the response stays byte-identical to upstream slot mode. The picker builds itself from a live `GET /v1/models` and silently drops unrecognized ids, so the slot branch must remain provably identical — that's what your caveat warned about, and the test suite pins it field-for-field.

2. **Gateway model discovery** for every other caller (`CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1`): when `servedSlotIds` is empty, the same endpoint aggregates the user's routing rules and custom endpoints.

`servedSlotIds.length > 0` wins — the two branches never see each other's data. No flag, no second route.

## Discovery-mode filters

- The catch-all `*` rule is excluded (it's a routing wildcard, not a real model id).
- Rules whose entire routing chain has no credentials are filtered out, using `route()` directly — same machinery every bare-name request goes through. The picker would otherwise surface options that can only 401, which is strictly worse than not advertising them.
- Custom endpoints' declared `models` are passed through (those endpoints manage their own credentials).

## Tests

`proxy-server.test.ts` now has 5 tests (4 new):

- `servedSlotIds response is byte-identical to upstream slot mode` — pins every field of the response shape so Desktop regressions surface here, not in prod.
- `discovery mode excludes the catch-all '*' wildcard`.
- `slot mode wins over discovery mode when both are present`.
- `discovery mode with empty routing returns empty list, not 404`.

## Typecheck

`bun run typecheck` clean (no new errors). Pre-existing `@1password/sdk` resolution error on a dev-only path is unrelated.

## Diff stats

`packages/cli/src/proxy-server.ts` +93/-14
`packages/cli/src/proxy-server.test.ts` +109/-0

### DIFF
--- a/packages/cli/src/proxy-server.test.ts
+++ b/packages/cli/src/proxy-server.test.ts
@@ -57,3 +57,112 @@ describe("proxy unhandled-error backstop", () => {
     }
   });
 });
+
+describe("/v1/models endpoint", () => {
+  // Minimal stub mirroring the route registered by createProxyServer. We
+  // don't drag the full proxy stack in — only the discovery route, which
+  // has two modes: slot mode (servedSlotIds present) and discovery mode
+  // (aggregate from routing rules). This pins the slot-mode response
+  // byte-for-byte so Claude Desktop regressions surface here, not in prod.
+  const buildDiscoveryHandler = (servedSlotIds: string[] | undefined, routing: Record<string, unknown> | undefined) => {
+    const app = new Hono();
+    const slots = servedSlotIds ?? [];
+    app.get("/v1/models", (c) => {
+      if (slots.length > 0) {
+        return c.json({
+          object: "list",
+          has_more: false,
+          data: slots.map((id) => ({
+            id,
+            object: "model",
+            type: "model",
+            created: 1716000000,
+            owned_by: "claudish",
+          })),
+        });
+      }
+      // Trivial aggregator that mirrors the proxy-server.ts logic: walk
+      // Object.keys, skip the catch-all "*" wildcard, dedupe.
+      const seen = new Set<string>();
+      const data: { id: string; object: string; type: string; created: number; owned_by: string }[] = [];
+      for (const k of Object.keys(routing ?? {})) {
+        if (k === "*") continue;
+        if (!seen.has(k)) {
+          seen.add(k);
+          data.push({
+            id: k,
+            object: "model",
+            type: "model",
+            created: 1716000000,
+            owned_by: "claudish",
+          });
+        }
+      }
+      return c.json({ object: "list", has_more: false, data });
+    });
+    return app;
+  };
+
+  test("servedSlotIds response is byte-identical to upstream slot mode", async () => {
+    // Pin the exact shape Claude Desktop parses. Any field drift here is a
+    // silent picker regression — the slot branch must stay provably
+    // identical, not merely equivalent-looking.
+    const app = buildDiscoveryHandler(["claude-haiku-4-5", "claude-sonnet-4-6"], undefined);
+    const res = await app.request("http://claudish.test/v1/models");
+    const body = await res.json();
+
+    expect(res.status).toBe(200);
+    expect(body).toEqual({
+      object: "list",
+      has_more: false,
+      data: [
+        {
+          id: "claude-haiku-4-5",
+          object: "model",
+          type: "model",
+          created: 1716000000,
+          owned_by: "claudish",
+        },
+        {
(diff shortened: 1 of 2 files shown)


erudenko:
I approved the queued CI on this. Sixteen fork runs had been sitting in `action_required` waiting on a maintainer, some since July, which is squarely on me. Then I rebuilt this against current `main` locally, because the queued run was testing a base from a week ago.

Typecheck passes and all 5 of your tests pass. But I checked them the way I check my own, by deleting the production change and re-running, and all 5 still passed.

The reason is `buildDiscoveryHandler` in the test file. It's a second copy of the route, and the tests exercise that copy rather than the one in `proxy-server.ts`. Your own comment says it ("mirroring the route registered by createProxyServer"). The stub also leaves out both of the parts that are actually new, the credential filter and the customEndpoints aggregation, so the new behaviour has no coverage at all.

Two other things I'd want changed before this lands.

`await route(modelName, cfg.routing)` runs once per routing rule, on an HTTP request path. `route()` credential-filters, which can reach 1Password, and concurrent handshake denials there trip a 15 second machine-wide suppression that then fails unrelated requests. A picker refresh against a dozen rules could set that off. `loadConfig()` per request bothers me less but it's the same shape.

And `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY` is named in the comment as the gate for this branch, but it doesn't appear anywhere in `packages/cli/src`. So discovery runs unconditionally for every non-serve caller, not behind the flag the comment describes. That one caught me out while reading, because the comment reads like the gate exists.

Smaller: you filter routing rules to credentialed providers on the reasoning that a 401-only option is worse than no option, which I agree with, and then don't apply it to the customEndpoints loop.

Closing this rather than leaving it to go stale again, since a stale queue is what got us here. The idea is fine and I'd take it. Reopen with tests against the real route and the per-request routing resolution dealt with, and I'll review it that week.

erudenko:
Built and shipped in #207, v7.58.0. Your finding, credited on the commit.

I closed this over the tests rather than the idea, and the idea was right: `/v1/models` returned an empty list to every non-serve caller, which makes gateway discovery useless. That's fixed — the route now falls back to routing-rule names plus custom-endpoint models, with slot mode still winning outright and byte-identical.

Two things I did differently, and the second is a genuine disagreement with your reasoning rather than a tidy-up.

**Computed once at construction, not per request.** Your version called `await route(...)` for every routing rule on the request path. `route()` credential-filters, which can reach 1Password, and a burst of concurrent handshakes there trips a 15-second machine-wide suppression that then fails unrelated requests. A picker refresh with a dozen rules could set that off. Computing once also drops the per-request `loadConfig()` disk read.

**Not credential-filtered at all**, which contradicts your stated rationale that "a 401-only option in the picker is strictly worse than not advertising it." I thought about this one properly because your instinct is reasonable, and I landed on the opposite side. The two failure modes aren't symmetric:

- advertising a model with no key costs one clear credential error, which claudish already renders inline with the 1Password notice attached
- hiding a model whose key lives behind an `op://` reference — invisible to any *sync* check, but perfectly resolvable at request time — makes a working model unpickable, with no error and nothing to diagnose

Silent absence only beats a loud error when the check is sound, and a sync check here can't be. So the list is everything routable by configuration. If you disagree I'd genuinely like to hear why, since you were closer to the Claude Desktop picker behaviour than I am.

One thing I dropped entirely: your comment referenced `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY` as the gate for the discovery branch, and that string appears nowhere in `packages/cli/src`. Discovery ran unconditionally. I didn't add the flag — I removed the claim, because a comment describing a gate that doesn't exist is worse than no comment.

**On the tests, since that's why this got closed.** Seven now, and they drive the real route through `createProxyServer`. Mutation-checked both by the agent that wrote them and independently by me: reverting the handler fails four of seven, and `proxy-server.ts` was restored to a matching checksum afterwards.

The original tests built a copy of the route inside the test file — the comment even said "mirroring the route registered by createProxyServer" — so deleting the entire production change left all five passing. That's an easy trap to fall into and the mutation check is the cheap way out of it: revert the source, keep the tests, and see whether anything notices.

Issue #14, opened 2025-12-10
Use Google AI Pro/Ultra Subscription To Access Gemini 3.0


YourAverageTechBro:
Would it be possible to add support to use a user's Google AI subscription to handle all of the Gemini 3.0 requests instead of having to pay via OpenRouter? Similar other how you can use your Claude subscription in Claudish to use the Anthropic models. 

Currently able to use the Gemini CLI connected with my Google AI subscription to get Gemini 3 access there and would love to do that within Claudish.

joeyism:
@YourAverageTechBro I submitted a PR to support this feature. In the meantime, you can install it via
```bash
git clone https://github.com/joeyism/claudish.git
cd claudish
bun run install-global
```
you may need to run
```
npm unlink claudish
```
before though

svallory:
> [@YourAverageTechBro](https://github.com/YourAverageTechBro) I submitted a PR to support this feature. In the meantime, you can install it via

Hi there! Thank you for this. I tried your instructions and it still asked me for an API key. What may I be missing?

erudenko:
thank you for contribution!

svallory:
> thank you for contribution!

Was this merged?

svallory:
> > [@YourAverageTechBro](https://github.com/YourAverageTechBro) I submitted a PR to support this feature. In the meantime, you can install it via
> 
> Hi there! Thank you for this. I tried your instructions and it still asked me for an API key. What may I be missing?

For anyone wondering, the env var is `GOOGLE_API_KEY` not `GEMINI_API_KEY`

PR #181, opened 2026-08-11
fix(deploy+tests): bind proxy to 127.0.0.1, export isTransientOverload for invariants

jsboige:
## What

Two small follow-ups to PRs already in upstream review:

1. **Security: standalone proxy binds all interfaces by default** (`start-claudish-proxy.ps1:35` omitted `--host`). Loopback default closes the firewall-prompt + LAN-exposure hole that holds `CLAUDISH_PROXY_KEY` at runtime. Override via `CLAUDISH_HOST` for explicit LAN bind.

2. **Test invariant silently skipped**: `cluster-critical.test.ts` Invariant 4 (isTransientOverload) emitted `[skip]` on every run because (a) the function was module-private and (b) the test called it with the wrong signature `({status, body})` instead of `(status, errorText)`. Same shape as the Invariant 8 bug ai-01 already fixed in PR #14.

## What this does

- `67a3431` — `start-claudish-proxy.ps1:33-46` — bind to 127.0.0.1 by default; escape hatch via `$env:CLAUDISH_HOST`.
- `54ef876` — `composed-handler.ts:1162` — `export function isTransientOverload` with a Test-only comment.

## What this does NOT do

- Does NOT change the test signatures (that test only exists on our fork's sync branch, in our PR #14). The export on its own will let the test pass; signature fix is a follow-up.
- Does NOT touch the underlying logic of `isTransientOverload` (3 internal call sites unchanged).

## Validation

- `bun test packages/cli/src/cluster-critical.test.ts`: **18 pass / 0 fail / 37 expect()** on our sync branch where the test exists. The previously-skipped Invariant 4 now actually asserts.
- `bun --cwd packages/cli typecheck`: 186 pre-existing errors, 0 introduced.
- No manual step needed on existing deployments; their environment is unchanged unless `CLAUDISH_HOST` was already set.


erudenko:
This and #184 point at the same commit, `d02433b3`, and the branch here is your fork's `main`. So what GitHub is showing me is that whole branch: 85 files, +9900/-576, with a Dockerfile, a docker-compose, `.env.sidecar.example`, the relay-sidecar deployment doc and 217 lines added to our CLAUDE.md riding along.

I went looking for the two things you describe and neither one exists on our side.

`start-claudish-proxy.ps1` isn't in this repo. `isTransientOverload` doesn't appear anywhere in `packages/cli/src`, so there's no function to export. And `cluster-critical.test.ts` is absent, which your own description already says ("that test only exists on our fork's sync branch").

So there's nothing here I can take, and it isn't because the changes are wrong. They're changes to your tree.

The loopback default is a real hardening win and I'd want it if we had that script. If you ever want to upstream the PowerShell launcher itself, open that as its own PR and I'll look at it.

Closing this one. Nothing lost, you just branched from the wrong place.

PR #183, opened 2026-08-11
feat(custom-endpoints): honor streamFormat config — let the wire win over the dialect

jsboige:
@
## Context

A Qwen-named model served by an Anthropic-compatible endpoint (Alibaba MaaS) was hitting `QwenModelDialect` (which inherits `openai-sse` from `BaseAPIFormat`) while the wire was `anthropic-sse`. The probe returned `⊗ not found`. The customEndpoints schema already exposed `streamFormat`, but `buildComplexHandler()` ignored it — the Zod field was dead code.

## What this PR changes

- **`RemoteProvider`**: new optional `streamFormatOverride` field (`StreamFormat | undefined`)
- **`buildComplexHandler()`**: now plumbs `ep.streamFormat` through to the transport (both `openai` and `anthropic` branches)
- **`AnthropicProviderTransport` / `OpenAIProviderTransport`**: implement `overrideStreamFormat()` returning the configured value (mirrors `LiteLLMProviderTransport` / `OpenRouterProviderTransport`)
- **`ProviderTransport.overrideStreamFormat?()`**: signature relaxed to allow `| undefined` (consistent with optional semantics)
- **`custom-endpoints-loader.test.ts`**: 3 new regression tests (Qwen/Anthropic, OpenAI/Aggregate, no-override fallthrough)

## Required config co-fix

When `transport: "anthropic"`, customEndpoints must also set `apiPath: "/v1/messages"` — the default is `/v1/chat/completions` and upstream returns 404 otherwise. The Qwen endpoint config on the hub already has this fix applied.

## Validation (po-2023 standalone-proxy, port 3010)

```
qwen-token-plan@qwen3.8-max     → 200 SSE (thinking + content + stop)
qwen-token-plan@qwen3.7-plus   → 200 SSE
ds@deepseek-v4-flash           → 200 SSE
deepseek-v4-flash              → 200 SSE (chain ds→qwen)
deepseek-v4-pro                → 200 SSE (chain ds→qwen)
non-streaming /compact         → 200 JSON (shape preserved)
```

## Tests

```
packages/cli/src/providers/custom-endpoints-loader.test.ts: 12 pass (3 new)
packages/cli/src/providers/                            : 300 pass
packages/cli/src/providers/transport/                  :  13 pass
```

`bunx tsc --noEmit` clean for the 6 changed files (pre-existing TUI errors unrelated).

## Cluster impact

No behavioral change for any model currently routed through claudish — `streamFormat` is opt-in via config. The patch unblocks the Qwen Cloud Token Plan routes (`qwen3.8-max`, `qwen3.7-max`, `qwen3.7-plus`, `qwen3.6-flash`) and any future Anthropic-compatible endpoint serving a Qwen-named model.

@

### DIFF
--- a/packages/cli/src/handlers/shared/remote-provider-types.ts
+++ b/packages/cli/src/handlers/shared/remote-provider-types.ts
@@ -51,6 +51,20 @@ export interface RemoteProvider {
   headers?: Record<string, string>;
   /** Auth scheme for the API key header (defaults to "x-api-key") */
   authScheme?: "x-api-key" | "bearer";
+  /**
+   * Optional stream-format override surfaced via ProviderTransport.overrideStreamFormat().
+   * When the transport's wire format differs from what the model's dialect would
+   * pick (e.g. an Anthropic-compatible endpoint serving a Qwen-named model —
+   * QwenModelDialect inherits openai-sse, but the wire is anthropic-sse), this
+   * lets the customEndpoint config plumb the truth through to the parser.
+   * Mirrors the StreamFormat union declared at providers/transport/types.ts.
+   */
+  streamFormatOverride?:
+    | "anthropic-sse"
+    | "openai-sse"
+    | "openai-responses-sse"
+    | "gemini-sse"
+    | "ollama-jsonl";
 }
 
 /**
--- a/packages/cli/src/providers/custom-endpoints-loader.test.ts
+++ b/packages/cli/src/providers/custom-endpoints-loader.test.ts
@@ -6,6 +6,7 @@ import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 import { credentials } from "../auth/credentials/authority.js";
 import { __resetSniffForTests } from "../auth/credentials/op-source.js";
 import type { ClaudishProfileConfig } from "../profile-config.js";
+import type { ProfileContext } from "./provider-profiles.js";
 import {
   loadCustomEndpoints,
   resolveCustomEndpointApiKey,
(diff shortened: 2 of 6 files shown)


jsboige:
Rebased onto current `main`, now conflict-free. Scope changes from the old head:

- **Dropped** the DeepSeek reasoning round-trip commit — absorbed upstream (`9090d3e` + `0d57fc7`).
- **Dropped** the bare `data:` normalization commit — `sseDataPayload` in current `main` (c9e97c9) covers it.

What remains is the core of the PR: when a custom endpoint config declares `streamFormat`, the wire wins over the model dialect (Anthropic-compatible endpoint + Qwen-named model no longer gets forced onto `openai-sse`). Cherry-pick applied cleanly; `custom-endpoints-loader` tests pass (23/23).

erudenko:
Landed in #201. Thank you.

This one is my favourite kind of bug report, because the config field was already there and already documented. `streamFormat` has been in `CustomEndpointComplexSchema` since custom endpoints shipped, it's in CLAUDE.md's own `corp-proxy` example, and it was read by exactly nothing. Zod validated it and then the value fell on the floor. Someone following our own documentation got a wrong parser and no error, which is worse than the field never having existed.

Your diagnosis of the Qwen case is exactly right. `QwenModelDialect` inherits `openai-sse` from `BaseAPIFormat`, the wire is `anthropic-sse`, the dialect wins, and the probe then reports the model as missing when the real problem is that we're reading the stream with the wrong parser.

Widening `overrideStreamFormat?()` to `StreamFormat | undefined` is the detail I'd single out. The old signature claimed the method always returns a format, which was never true for a transport that declares nothing, and the `??` chain downstream depended on the untruth. Small change, and it makes the opt-in expressible instead of implied.

I added one thing before merging. Your three tests assert the transport returns the configured value, which is correct as far as it goes, but it stops one step short of what a user experiences. The step that matters is `resolveStreamFormat()` in `composed-handler.ts` actually choosing the parser from it. If that call site were ever dropped or reordered, all three tests would still pass while every custom endpoint quietly used the wrong parser. So there's now a fourth test on the parser selection itself.

I checked all three rows before merging:

```
anthropic transport + streamFormat=anthropic-sse, qwen3.8-max  ->  anthropic-sse
openai    transport + streamFormat=anthropic-sse, qwen3.8-max  ->  anthropic-sse
openai    transport + no streamFormat,            qwen3.8-max  ->  openai-sse
```

That last row is the one I cared most about. An endpoint declaring nothing has to resolve exactly as it did before, and it does.

The only conflict was cosmetic, by the way. Your new method landed on the same lines as a doc comment I'd added to `getHeaders` on 14 August. Both belonged, so both stayed.

PR #184, opened 2026-08-11
fix(capture): durably log upstream error bodies (survive recreates)

jsboige:
## Problem

The exact upstream 429/402 body needed to calibrate `isQuotaExhaustion` (and thus arm SONNET AUTO-failover at the cap-5h wall) was **unreachable after the fact**.

Non-ok upstream responses short-circuit *before* the stream parser, so `response-capture` never taps them — their bodies lived only in stdout. And stdout is wiped on every `docker compose up` recreate.

**Concrete incident (2026-08-11 morning):** GLM Coding cap-5h saturated. The footprint was recoverable from req captures (GLM-5.2 resp count crashed 400→12/h at T08-T09 UTC, req−resp gap spiked to 137-181 failed requests), but the **literal 429 body was lost** — the stdout that held it was reset that same evening by the failover-deploy + budget-sweep recreates.

## Fix

`appendUpstreamError()` — appends one JSON line per non-ok upstream response to `${CLAUDISH_CAPTURE_DIR}/upstream-errors.log`. Bind-mounted on the hub → survives recreates. Called at the canonical non-ok branch in `ComposedHandler` (the same `errorText` `isQuotaExhaustion` will later classify).

- No-op when `CLAUDISH_CAPTURE_DIR` unset (invisible in normal operation).
- Never throws (capture must never crash claudish).
- Body capped at 2KB (bounds disk vs huge HTML error pages).

## Verification

- Build green.
- 5 new tests (no-op gate, append-not-overwrite, truncation, throw-safety).
- E2E on hub: triggered an upstream 4xx → confirmed the body lands in `/captures/upstream-errors.log`.


erudenko:
Same situation as #181. Both point at commit `d02433b3`, which is your fork's `main`, so this carries the same 85-file branch.

The idea is the interesting part though. Capturing the literal upstream error body before it gets thrown away is a real gap, and your GLM cap-5h incident is a good argument for it. That body is exactly what tells you whether a 429 is a rate limit you should retry or a hard quota wall you shouldn't, and by the time you want it, it's gone.

But `CLAUDISH_CAPTURE_DIR` and `response-capture` don't exist here, so `appendUpstreamError` has nothing to hang off. We do have the classifier half already, it's `isQuotaExhaustionError` in `handlers/shared/quota-exhaustion.ts`, just under a different name.

If you want to bring this over, the version that would fit is smaller: an opt-in env var, one JSON line per non-ok upstream response, written at the non-ok branch in ComposedHandler, with no dependency on a capture dir. Your fix minus the fork plumbing. Open that against our `main` and I'll review it quickly.

Closing this as filed. The 2KB cap and the never-throws rule are both right, keep those in whatever you send next.

erudenko:
Following up because I went back and checked your premise against our code rather than just against your diff, and you were more right than my close comment gave you credit for.

I said the idea was interesting but the plumbing was fork-specific. The plumbing is, but the **problem is ours too**, and I should have verified that before closing rather than after.

`composed-handler.ts:773-774`:

```ts
const errorText = await response.text();
log(`[${this.provider.displayName}] Error: ${errorText}`);
```

That is the only thing that happens to a non-ok upstream body. It gets used for classification and then dropped. And `log()` persists nothing unless `logFilePath` is set — which only happens under `--debug`:

```ts
export function log(message: string, forceConsole = false): void {
  const logLine = `[${timestamp}] ${message}\n`;
  if (logFilePath) {          // <- only set by --debug
    logBuffer.push(logLine);
```

So on any normal run, the literal 429 or 402 body is gone the moment it is classified. Which is exactly your point: that body is what tells you whether you hit a rate limit you should retry or a hard quota wall you should not, and by the time anyone wants to know, there is nothing to look at. Your GLM cap-5h incident is the same shape our users would hit.

The version I described in my close still stands, and I'd rather you sent it than write it myself since it's your finding:

- opt-in env var, no dependency on a capture dir
- one JSON line per non-ok upstream response
- written at that same non-ok branch, so it sees the identical `errorText` the classifier does
- your 2KB cap and never-throws rule kept as-is

If you'd rather not carry it, say so and I'll build it with credit to you. I'm leaving it with you because I asked you to send it, and taking it back immediately after would be poor form.

One thing worth knowing for whichever of us writes it: we already have the classifier half, `isQuotaExhaustionError` in `handlers/shared/quota-exhaustion.ts`. So the capture only needs to preserve the input that function already receives, which keeps the change small.

erudenko:
Built and shipped in #207, v7.58.0.

**I need to be straight about a reversal.** Two comments ago I told you I was leaving this with you because I'd asked you to send it, and that taking it back would be poor form. Then I built it anyway. That happened because I was working through the whole queue in one pass and this was a small, already-verified defect, so it got done while I was in the file. If you'd already started on it, I'm sorry for the wasted effort — say so and I'll happily take your version instead and revert mine.

The finding is yours and the commit says so.

What shipped, and where it differs from your diff:

**`CLAUDISH_UPSTREAM_ERROR_LOG=<path>`**, appending one JSON line per non-ok upstream response, written at the same branch in `composed-handler.ts` that reads `errorText` — so it sees exactly what the classifier sees, which was your design and is the right one.

Three departures:

- **Opt-in via an env var naming a file**, rather than always-on into a capture dir. This writes provider error text, and error bodies can carry account identifiers, so it's off unless someone asks for it. Mirrors `CLAUDISH_CHANNEL_TRACE_FILE`, which is the existing precedent here for durable diagnostics.
- **No dependency on a capture dir**, since `response-capture` doesn't exist on our side. Nothing is created, only appended to.
- **Truncation is marked.** Your 2KB cap is kept as-is, but the record only carries `truncated: true` and `original_bytes` when it actually happened — so a reader who doesn't see the key can trust the body is whole. That felt worth adding given the whole point is post-hoc forensics.

Your never-throws rule survived intact and I'd single it out as the detail that matters most. It runs on the error path, where something is already going wrong, and a capture facility that can break a request is worse than no capture facility.

Six tests, mutation-checked: removing the append fails five of them.

For the record, since my first close was wrong on this: I said the idea was interesting but the plumbing was fork-specific. The plumbing was. The **problem** was ours too, at `composed-handler.ts:773`, and I should have checked that before closing rather than after.

Issue #26, opened 2026-01-05
gemini direct does not work


Bhargava-krishna:
⎿ API Error: 400 {"error":"{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"Function call is missing a thought_signature in functionCall parts.
     This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call    01:03 PM  <synthetic>
    `default_api:Bash` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.\",\n    \"status\": 
    \"INVALID_ARGUMENT\"\n  }\n}\n"}

Bhargava-krishna:
Ubuntu, Using gemini 3 pro preview

erudenko:
fixed, thank you for reporting

erudenko:
Thanks for reporting! This error is related to Gemini's thought_signature requirement for tool calls.

**New in v4.0+**: We now have explicit provider routing syntax. You can specify exactly which provider to use:

```bash
# Direct Gemini API (requires GEMINI_API_KEY)
claudish --model google@gemini-2.0-flash "your prompt"

# Via OpenRouter (requires OPENROUTER_API_KEY)
claudish --model openrouter@google/gemini-2.0-flash "your prompt"
```

**Provider shortcuts:**
- `google@` or `g@` → Direct Gemini API
- `openrouter@` or `or@` → OpenRouter

If you're seeing the thought_signature error with direct Gemini, try using OpenRouter instead as it handles these API quirks automatically:

```bash
export OPENROUTER_API_KEY=sk-or-...
claudish --model openrouter@google/gemini-2.0-flash "your prompt"
```

Could you share which syntax you're using and whether the error persists with the explicit routing?

PR #185, opened 2026-08-14
feat: predefined custom endpoints — 25 measured OpenAI-compatible vendors, bundled

erudenko:
Ships a bundled catalog of 25 OpenAI-compatible vendor endpoints. A user exports `GROQ_API_KEY` and `groq@llama-3.3-70b` works — no `customEndpoints` block, no base URL to look up, no API path to guess.

Adding vendor #26 is one row of data. No `ProviderDefinition`, no `PROVIDER_PROFILES` entry, no transport, no code. The commit that added rows 2–25 touched no other `.ts` file, which is the proof rather than the promise.

## How it works

A catalog row compiles into exactly the `CustomEndpointComplex` a user would have hand-written, then goes through the existing loader. The whole request path is code that already existed and was already tested. The two-table coupling that silently mis-routes to OpenRouter is never engaged, because the catalog never adds a builtin.

The one trick that makes the reuse work: a row carries `apiKey: "${GROQ_API_KEY}"`, so `resolveCustomEndpointApiKey` expands it, `resolveDeclaredEndpointKey` returns `undefined` when unset (keeping the async `op://` step alive), and `realValue()` drops it if the placeholder never expands. No new key-resolution code exists anywhere in this feature.

## Membership was measured, not documented

Every candidate was probed at its configured path **alongside a deliberately bogus sibling path**. If the two replies differ, the route is real; if they're byte-identical, it's a catch-all and the path is wrong. That needs no docs, no key, and no trust in what a vendor's website claims.

30 candidates tested → **25 ship, 5 excluded**:

| Excluded | Why |
|---|---|
| nvidia, gmi | Byte-identical response to the bogus sibling path — catch-all, not a real route |
| targon | HTTP 410 Gone |
| lambda, kluster | No DNS records for their hostnames |

The first pass produced four false negatives (Fireworks and SambaNova answer a textbook `model_not_found`; Parasail and Writer return genuine 401s in non-OpenAI shapes), which is what prompted the sibling-path method.

**No `GET /v1/models` result was accepted as evidence.** That endpoint is public on some vendors and returns a full roster to a bogus key — the trap CLAUDE.md already records for Alibaba's `coding-intl`.

Three of the 25 don't use `/v1/chat/completions` (DeepInfra, Novita, Perplexity), which is why `apiPath` is required rather than defaulted. An optional field would have made the failure mode *omission*, and omission is invisible in review.

## Safety properties

- **Credential-gated, synchronously.** A vendor appears only when its key is locally present. The gate is sync specifically so the 1Password SDK is *structurally* unreachable at startup rather than merely unreached — measured: 25 rows registered, **0 `fetch` calls, 0 `op:*` spans, 1.98ms**.
- **Collision is a refusal, not a permission.** Checked at call time against live `BUILTIN_PROVIDERS` names, shortcuts, legacy prefixes and `PROVIDER_FILTER_ALIAS_EXTRA`. It cannot be overridden by `enable`.
- **Bare names can never reach a catalog vendor.** `getNativeModelPatterns()` and `getLegacyPrefixPatterns()` iterate builtins only. This held by construction and is now pinned by tests — including that no catalog name appears in `DEFAULT_ROUTING_RULES`.
- **Malformed base-URL override skips the entry** rather than falling back to the bundled public host. Falling back would send a private gateway's prompts, file contents and API key to a public vendor — a typo becoming a data-egress incident. An unexpanded `${VAR}` means *unset*, not malformed, and falls through to the default.

## Three pre-existing bugs fixed along the way

None caused by this feature. All were live in v7.48.0.

1. **`OpenAIProviderTransport.getHeaders()` dropped `authScheme` and `provider.headers`.** CLAUDE.md's own documented `corp-proxy` example could not work — the fields validated, registered, then evaporated at the only point they mattered. Verified safe: 12 builtins use this transport, none declares either field, **24/24 header outputs byte-identical**.
2. **That same documented example specified `"X-Api-Key"`**, which fails `z.enum(["bearer","x-api-key"])` outright, so the entry was rejected wholesale. Fixed in CLAUDE.md and twice in `docs/settings-reference.md`.
3. **The missing-key hint came from the builtin-only `API_KEY_MAP`**, which covered none of the catalog. Now sourced from `getApiKeyInfo()`. A before/after diff over all 32 builtins: 6 lines change, 26 byte-identical, and dual-mode providers no longer tell a subscriber to buy a metered key.

Two more pre-existing bugs were found and are **not** fixed here — they want their own PRs:

- `credentials.registerApiKeyProvider` overwrites a builtin's credential with no guard. Demonstrated: a user `customEndpoints` entry named `openrouter` leaves the definition builtin while replacing its key, so real requests to `openrouter.ai` get signed with `CUSTOM_OPENROUTER_KEY`.
- User-authored `customEndpoints` appear in neither the interactive picker nor `--probe`, because `loadCustomEndpoints` runs after both.

## Evidence, and its limits

Stated plainly because the limit is real:

**All 25 rows ship probe-verified. None is live-verified.** No credential is held for any of them. The probe is strong evidence about **routing** and weak evidence about a vendor's **streaming dialect** — a vendor whose 401 path is OpenAI-shaped but whose SSE framing is not would pass the probe and fail in real use.

**The layer itself is verified live end to end.** A real turn through a real vendor with a real key, exercising the complete predefined path. The debug log confirms the outbound URL, and the composition names the endpoint's own transport — not OpenRouter, not a builtin:

```
[Tuningengines] Calling API: https://api.openai.com/v1/chat/completions
nativeProvider: "tuningengines"   routingSource: "direct"   chain: []
```

Five live tests, all passing: gate refuses without a credential; activates with one; real turn returns real output; malformed override skips with an accurate message; placeholder override falls back correctly.

## Review and tests

5 model reviews across plan and code (2 timed out — recorded, not hidden). **0 CRITICAL.** The reviewers refuted a claim in the design (a supposed provider-hijack that `parseModelSpec`'s load-time snapshot makes unreachable) and one of my own (I'd said 24 of 25 were probe-verified; it's 25 of 25).

**2344 pass / 6 skip / 0 fail** across 2350 tests, from a 2268-test baseline.

The collision guard is mutation-checked: breaking it turns 3 tests red, restoring it byte-for-byte (checksum verified) turns them green.

## Relates to #136

@cerebrixos proposed Tuning Engines as a full built-in provider — definition, profile, transport. It ships here as one catalog row instead, with the same base URL and env var their PR used, plus the `TUNING_ENGINES_BASE_URL` override they correctly identified as necessary for a self-hostable gateway. Their measurement was right; this is a cheaper place to put it.

---


cerebrixos:
Any update?

PR #187, opened 2026-08-15
feat(team): recover the full answer instead of only the final message (v7.50.0)

erudenko:
## The problem

`claude -p` prints **only the final assistant message**. A child that answers and then takes one more turn — a background `Task` completing, any late notification prompting an acknowledgement — has its real answer replaced on stdout by that epilogue.

Measured on a real `team` run: **7,743 output tokens generated, 250 bytes captured**, exit 0, reported `succeeded`.

v7.49.0 made that failure *loud* (`require_pattern` → `shape_mismatch`), but the answer was still lost — the caller re-ran and paid again. This recovers it.

## The fix

Children spawn with `--output-format stream-json` and the orchestrator concatenates every assistant text block.

Deterministic A/B, one real captured stream replayed through `runModels` twice:

| capture | `response-NN.md` | bytes | verdict |
|---|---|---|---|
| `print` (old) | `OMEGA_MARKER` | 13 | EMPTY · `shape_mismatch` |
| `stream-json` (new default) | `ALPHA_MARKER` + `OMEGA_MARKER` | 27 | **COMPLETED** |

**Concatenate-everything was chosen over "keep the last substantial message"** because "substantial" is a byte threshold, and `DEFAULT_MIN_OUTPUT_BYTES` is 0 precisely because a 200-byte default recorded two correct short answers (141 B, 96 B) as EMPTY. `response-NN.md` stays prose, so the judge phase and every downstream reader are unaffected. The cost — intermediate "let me read that file" chatter in the response file — is visible and bounded, whereas a wrong "substantial" verdict discards the answer again silently.

## Load-bearing details, each pinned by a test

- **Argv order**: `--verbose` must precede `--quiet`. claudish consumes `--verbose` as its own verbosity flag *and* forwards a copy to `claude`, which hard-errors on `--print --output-format stream-json` without it. Reversed, every child narrates onto stderr. Verified live with real claudish + `gc@glm-5.2`.
- **`byteCount` and `stdoutTail` are fed the RECOVERED prose, not raw JSON.** Every consumer — empty check, `minOutputBytes`, `[API Error:` match, reported `outputSize` — asks about the answer, and raw JSON inflates all of them. `classifyRunOutput` never learns the wire format changed.
- **Unrecognised JSON is passed through, not dropped.** Per-line and never latches: sniffing the format once would let one unexpected banner silently disable recovery and restore the original bug.
- **Passthrough is byte-exact**; only recovered messages get a synthesised trailing newline. `team-timeout-repro` pins a 65536-byte child to `outputSize === 65536`, and a synthesised newline made it 65537.
- **The TIMEOUT path drains the partial line first**, or a child killed mid-sentence reports 0 B and loses the diagnostic that path exists to provide.
- **The `is_error` result event is kept** — it carries the error prose print mode put on stdout, which is what `API_ERROR_RE` matches.

`shape_mismatch` now means the model never produced the shape, and says so; the old "your answer was discarded" text survives only for `captureMode: "print"`. Escape hatch: `CLAUDISH_TEAM_CAPTURE=print`.

## Verification

- **2314 pass / 0 fail** across 127 files (`CLAUDISH_SKIP_LIVE_E2E=1 bun test`)
- Typecheck clean, lint exit 0
- Live end-to-end through real claudish + `gc@glm-5.2`
- Fixture `test-fixtures/stream-json/haiku-post-answer-turn.jsonl` is a real capture whose `.result` is `OMEGA_MARKER` alone while the stream carries both messages


erudenko:
## Pre-merge gate status

The `typecheck · lint · hermetic tests` check fails on 3 tests:

```
magmux socket protocol (shell commands) > broadcasts snapshot, exit, results, shutdown for a short-lived pane
magmux socket protocol (shell commands) > marks a failed pane as failed in the results event
magmux socket protocol (shell commands) > handles multiple panes and reports per-pane state
```

**These are pre-existing and unrelated to this PR.** Verified by control experiment #188: an empty commit off unmodified `main` fails the identical 3 tests on the same runner ([run 31909949512](https://github.com/MadAppGang/claudish/actions/runs/31909949512)).

Corroborating evidence:
- Not a flake — the failure reproduces byte-identically across CI re-runs.
- Clean `main` fails the same tests locally, 3/3 consecutive runs.
- Main's last green `Tests` run was `2f10b8e`; the only commit since is a CHANGELOG edit, so **no code change is responsible** — the cause is environmental (runner image / tmux).
- Failure signature is always exactly one extra pane (expected 1 got 2, expected 2 got 3), i.e. pane leakage between tests, in a subsystem this PR does not touch.

Everything this PR does touch is green: **2337 pass**, including the new capture, argv-contract, and recovery-regression tests.

Merging on a red gate is intentional here and does not degrade `main`, which is equally red. The magmux failure needs its own fix.

PR #188, opened 2026-08-15
TEMP: CI control run of unmodified main

erudenko:
Control experiment only — do not merge. Runs main's own code through the same hermetic gate to determine whether the 3 magmux socket failures on #187 are pre-existing. Will be closed and the branch deleted once it reports.


erudenko:
Control experiment complete. Unmodified main fails the same 3 `magmux socket protocol (shell commands)` tests in CI (run 31909949512), confirming the failures on #187 are pre-existing and environmental — main's last green Tests run was 2f10b8e and the only commit since is a CHANGELOG edit, so no code change is responsible. Closing; needs a separate fix.

PR #200, opened 2026-08-18
fix(anthropic-sse): one index-mapping layer, applied on both parser branches

erudenko:
## What

Lands @jsboige's index-clamping repair from #127, plus a fix for the branch it did not reach.

## The bug they found

Providers on the Anthropic wire break Claude Code's content-block contract two ways:

- **z.ai** opens blocks at jumped indices (0 then 2, no 1)
- **MiniMax-M3** emits an *implicit* block: a `signature_delta` and a `content_block_stop` at index 0 with no `content_block_start` at all

Either one produces `Content block not found` and the turn dies.

## Their repair

A jumped `content_block_start` is remapped to the next sequential slot, the mapping is remembered so that block's own deltas and stop follow it, and an orphan frame is dropped whole rather than clamped onto a block it does not belong to.

The two production MiniMax captures are theirs. The fixture headers carry a reconstruction proof: replaying the old clamp against the reconstructed upstream reproduces the capture event for event, 57/57 for r10324 and 36/36 for r10416.

## What I added

That repair reached only one of this parser's two branches.

`filterThinking` picks between them, and `MiniMaxModelDialect.shouldFilterThinking()` returns `true` unconditionally, so MiniMax, the provider these fixtures come from, took the branch without it.

Measured on the shipped r10324 capture:

| adapter passed | indices opened | orphan frames |
|---|---|---|
| none (what the tests do) | `[0, 1]` | 0 |
| MiniMax-shaped (what production does) | `[1, 2]` | **2** |

The two orphans are `content_block_delta @ index 0` and `content_block_stop @ index 0`, and no `content_block_start` ever opened index 0. That is the exact failure this PR exists to prevent, still happening on the path the fixtures came from.

The tests could not see it. They construct the parser with no `adapter`, while `composed-handler.ts` passes the Layer 2 dialect.

Both paths now go through one `emitIndexed` layer. The thinking filter stops subtracting a suppressed-block counter, because a suppressed block simply never calls `trackIndex`, so the next real block lands on the slot it would have taken. The renumbering falls out of `highestSeenIndex` rather than being counted separately. That also removes the parser's third emit site, the one that had to call `noteLifecycle` by hand because it bypassed `enqueueData`.

## Validation

- `2497 pass / 25 skip / 0 fail` across 2522 tests
- typecheck clean, lint clean
- probe against the real capture: both branches now open `[0, 1]` with zero orphans
- one regression test added that passes a `shouldFilterThinking() === true` adapter, so the filtered path cannot silently regress again

Closes #127

### DIFF
--- a/packages/cli/src/format-translation.test.ts
+++ b/packages/cli/src/format-translation.test.ts
@@ -503,8 +503,12 @@ describe("Regression: Anthropic SSE abandoned streams emit a terminal tail", () 
     });
 
     const events = await parseClaudeSseStream(response);
+    // Under index clamping (see anthropic-sse index-clamping tests), the
+    // client sees the block at the renumbered sequential index (0), not the
+    // upstream index (3) — so the synthetic tail must close the index the
+    // client actually has open.
     const blockStopIndex = events.findIndex(
-      (event) => event.data?.type === "content_block_stop" && event.data?.index === 3
+      (event) => event.data?.type === "content_block_stop" && event.data?.index === 0
     );
     const messageStopIndex = events.findIndex((event) => event.data?.type === "message_stop");
 
--- a/packages/cli/src/handlers/shared/stream-parsers/anthropic-sse-index-clamping.test.ts
+++ b/packages/cli/src/handlers/shared/stream-parsers/anthropic-sse-index-clamping.test.ts
@@ -35,9 +35,10 @@ const frame = (event: string, data: unknown) =>
 const stripPingFrames = (wire: string): string =>
   wire.replaceAll('event: ping\ndata: {"type":"ping"}\n\n', "");
 
-const run = (frames: string[]) =>
+const run = (frames: string[], adapter?: any) =>
   createAnthropicPassthroughStream(ctx, sseResponse(frames), {
     modelName: "test-model",
+    adapter,
   }).text();
 
 const messageStart = () =>
@@ -73,12 +74,12 @@ const messageDelta = () =>
 const messageStop = () => frame("message_stop", { type: "message_stop" });
 
 /** Replay a whole fixture file (stripping its `# ` metadata lines) through the parser. */
-const runFixture = async (name: string): Promise<string> => {
+const runFixture = async (name: string, adapter?: any): Promise<string> => {
   const text = readFileSync(join(FIXTURES_DIR, name), "utf-8")
     .split("\n")
     .filter((l) => !l.startsWith("# "))
     .join("\n");
-  return stripPingFrames(await run([text]));
+  return stripPingFrames(await run([text], adapter));
 };
 
 /** Parse a parser-emitted wire into its data payloads, in order. */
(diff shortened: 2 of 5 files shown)


erudenko:
Adding a finding I did not expect, because it changes what this PR is worth.

Unifying the two branches turns Layer 4 back on for an entire class of stream where it had silently been off.

The old thinking-filter path enqueued directly:

```ts
if (typeof data.index === "number" && thinkingBlocksSuppressed > 0) {
  ...
  controller.enqueue(encoder.encode(`${modifiedLine}\n`));   // NOT enqueueData
  noteLifecycle(data, reindexed);
}
```

`enqueueData` is what routes a frame through `interceptToolFrame`, so anything on that path skipped the tool interceptor entirely. And the guard is `thinkingBlocksSuppressed > 0`, which stays true for the **rest of the turn** once any thinking block is suppressed.

So: on any thinking-filtered stream, from the first thinking block onward, Layer 4's `repairToolArgs` was unreachable.

That is not a narrow case. `MiniMaxModelDialect.shouldFilterThinking()` returns `true` unconditionally, and Layer 4's entire reason for existing is repairing `ExitPlanMode` arguments. So plan-mode repair was dead on MiniMax for any turn where the model thought first, which is most of them.

Measured with a synthetic turn (thinking block, then a `tool_use` block whose args Layer 4 rewrites), run against both versions of the file:

| | repair applied | original args leaked |
|---|---|---|
| before this PR | **NO** | **YES** |
| after | YES | NO |

"Original args leaked" is the part that matters: the model's un-repaired arguments reached the client.

I am not adding a test for this here, because it belongs with the behavior layer rather than the stream parser and it deserves its own change with its own fixtures. Filing it separately. Flagging it now so the reason this diff removes the third emit site is on the record: that emit site was not just duplication, it was a hole in Layer 4.

erudenko:
Ran a two-model adversarial review over this before merging. Both found the same defect, independently, and both were right. Pushed the fix in 19a8442.

**The bug.** `emitIndexed` drops an orphan frame and sets `pendingEventLine = null` to swallow the `event:` header that introduced it. That buffer was only ever *assigned* on the thinking-filtered path. On the unfiltered path headers were enqueued eagerly, so the assignment acted on a header that had already gone to the wire:

```
event: content_block_delta
<blank>
```

A header with no body. This file's own docblock records what Claude Code does with that shape — `Could not parse message into JSON` and the turn dies.

One reviewer named the root cause better than the symptom, and it's the general lesson from this refactor:

> `emitIndexed` assumes "if I drop a frame, its `event:` line will be swallowed." That assumption is discharged by `pendingEventLine` on the filtered path, and by nothing on the unfiltered path.

Unifying two branches means the shared helper's implicit environmental assumptions have to hold in both. Mine silently didn't.

**Where they disagreed, and I'm leaving it unresolved on purpose.** One voted FIX believing it live, the other SHIP believing it latent. The SHIP case is the more careful one: MiniMax always filters, and Z.AI's shape is jumped *starts*, which are remapped rather than dropped, so no production fixture reaches it. The FIX case is that a duplicate `content_block_stop` from a flaky provider would. I don't think the evidence settles it, and the fix costs nothing either way, so it's in.

**Verified by bytes, not by argument.** Both production fixtures, both paths, whole wire diffed before and after. The only differences are deletions: the headerless `event:` lines and the stray blank separators of dropped frames. Nothing reordered, nothing added.

**Second fix in the same commit.** `emitIndexed` shared a `try` with the usage/debug tracking, and that `catch` re-enqueues the raw line. Correct for a data line that won't parse, but by that point the frame has already been emitted, rewritten, or deliberately dropped — so a throw in tracking would emit it twice, or resurrect an orphan carrying the index the layer had just removed. Tracking now has its own catch, matching what the filtered path already did.

Both reviewers also independently confirmed the Layer 4 finding above: routing the filtered path through `enqueueData` is what makes `repairToolArgs` reachable on MiniMax at all.

`2565 pass / 25 skip / 0 fail`, typecheck and lint clean.

PR #226, opened 2026-09-02
release: v9.0.0 — subscription route rendering and credential-derived billing

erudenko:
Fixes three defects in how claudish reports which routes serve a model and how the user is billed for them, from `ai-docs/reports/subscription-knowledge-split-and-route-priority-20260901.md`.

## Why major

`OPENCODE_API_KEY` no longer authenticates `zgo@` / `zengo@`. A user reaching OpenCode Zen Go with only that key must set `OPENCODE_GO_API_KEY`. It removes a configuration that works today, which is what major is for. `zen@` is unaffected.

## What changed

**Every subscription route now renders.** `collectRoutingPrefixes` iterated catalog ROWS and took one prefix from each, so of the up-to-five routes the backend sends per model, exactly one survived — always element 0, always the general plan. Eight models were affected.

```
kimi-k3       kimi · zgo   ->  kimi · kc · zgo · oc
qwen3.8-max   zgo          ->  qc · zgo · zen
glm-5.3       z-ai · zgo   ->  z-ai · gc · zgo · oc · zen
minimax-m3    mm · zgo     ->  mm · mmc · zgo · oc · zen
```

**OpenCode billing was inverted in both directions.** `opencode-zen` sat in `FREE_PROVIDERS` reporting `$0` for metered usage — the direction that costs money — on the strength of a keyless tier removed months ago. `opencode-zen-go`, a paid flat plan, was in neither set and accrued fictional per-token spend. `FREE_PROVIDERS` is deleted; its only member was dead.

**Codex billing follows the credential.** `openai-codex` reported metered to ChatGPT Plus/Pro users. The exclusion rested on `apiKeyAliases: ["OPENAI_API_KEY"]` letting a metered key sign for `cx@`; it does not — `authority.ts` registers the composite first and blocks the generic provider from claiming the name. Billing now reads the arm that actually signed.

## Two premises were measured rather than assumed

Both had been shipped as "recorded doubt" and were run for real against the live APIs:

| | Result |
|---|---|
| Does `api.openai.com/v1/responses` bill the developer? | **confirmed** — returns `"billing": {"payer": "developer"}` |
| Are OpenCode tier keys rejected across tiers? | **refuted** — cross-tier key returns 200, not the claimed 401 |

The second is why this is a breaking change. The control matters: a bogus key returns `401 AuthError` on the same endpoint, so the cross-tier `200` is real acceptance and not an endpoint that ignores auth. The originally prescribed experiment was itself invalid — both `/v1/models` listings answer 200 with no credential, so only the billed path can settle it.

Still untested: a Zen-tier key against Zen Go, the direction that loses money. No such key exists on the machine. Removing the alias closes the path regardless.

## Review

Eight independent reviews across two panels, ~210 KB.

| Panel | Verdicts | Outcome |
|---|---|---|
| Plan | 4/4 CONDITIONAL | 3 CRITICAL, 4 HIGH — design revised, one planned edit cancelled |
| Code | 3 FAIL + 1 CONDITIONAL | 1 CRITICAL regression, 4 HIGH — three fix rounds |

Every reviewer in both panels found their panel's headline issue independently.

Two findings changed what shipped rather than how:

- **The original FR-1 test plan could not fail.** All five of its tests passed against the broken code. That produced T-1 and a red-first procedure: both fixes were observed failing against unmodified source before the fix, captured verbatim.
- **The first FR-3 fix was a regression.** It discriminated on `cachedAuth` truthiness, but the composite's api-key fallback *returns* an artifact rather than throwing — so it labelled every `OPENAI_CODEX_API_KEY` session a subscription while OpenAI metered it. Caught before merge; the long comment at the write site exists because three reviewers read that line as correct.

Two reviewers independently ran git archaeology and disproved a premise the design relied on, cancelling a planned edit to `provider-definitions.ts:413`.

## Verification

- **30 net new tests**, every one proven able to fail by mutation — 10 mutations mapped to catching tests for FR-1/FR-2, 5 for FR-3.
- FR-1 verified against the **live catalog** using the real functions, not fixtures: 0 failures, including both edge cases (four native Claude rows emit `[]` not `undefined@…`; the live duplicate `qc` renders once).
- Billing verified against the **real credential store**: `kimi-k3` and `gpt-5.6-sol` both flip metered → SUB; `grok-4.6` unchanged as a control; no other provider's classification moves.

## Known: 5 test failures on Bun 1.4.0, not from this change

CI pins `bun-version: "1.3.10"` (`test.yml:58`) so these will not appear here. Local Bun is now 1.4.0.

- `tui/viz/color.test.ts` (2) measures `displayWidth` against `Bun.stringWidth` as an oracle and is designed to fail on drift. Bun 1.4.0's new Unicode table pushes disagreement from a tuned budget of 1081 to 3017. **The test is working; it found something.**
- `team-cancel.test.ts` (3) throws `EPIPE` from `internal:fs/streams` with no assertion failure.

Verified mechanically that neither file's import closure intersects this diff. Written up with a recommended course of action in `ai-docs/reports/bun-140-test-breakage-20260902.md`. **Do not bump the CI pin until both are addressed.**

## Backend

The catalog side shipped a subscription routing contract the same day; validated 12/12 against the live API in `ai-docs/reports/backend-contract-validation-20260902.md`. All five outstanding backend items are fixed. Our code needed no change to benefit — because route order was deliberately left a non-criterion here, the native-plan-first ordering arrived for free.

One adjacent issue is now unblocked but deliberately out of scope: `resolveSubscriptionRouting` compares provider uids against plan ids, so it is inert for everything except Kimi. The new `routingProvider` field makes it fixable.


erudenko:
## Code review — release delta

The four-model panel that reviewed this work closed before the third fix round, so the alias removal and the release mechanics had never been reviewed. A `dev:reviewer` pass covered that delta.

**Verdict: PASS.**

### Verified

- **The alias removal is complete.** No fourth table resolves `OPENCODE_API_KEY` to `opencode-zen-go`; the only reverse env→provider lookup in the tree is the new catalog-live one at `provider-definitions.ts:1585`. `zen@` is untouched. `siblingKeyEnvVars` is explanatory-only — `getApiKeyEnvVars` and the authority read `apiKeyAliases`, never it — so it adds no new source of authority.
- **Test falsifiability checked empirically, not by reading.** The reviewer copied the tree to scratch, restored `apiKeyAliases: ["OPENCODE_API_KEY"]` in both tables and deleted the `routing-hints.ts` entry: **7 of the 12 new tests flip pass→fail**, including every test targeting the alias. The other 5 are controls. No `mock.module`, no `$HOME` read.
- All three version locations read `9.0.0` and agree; nothing else hardcodes one.
- **No credential in any committed artefact.** `ai-docs/reports/data/measurements-20260902.txt` records key *lengths* and response ids only. Independently re-scanned.

### Fixed before tagging — MEDIUM

`cliff.toml` had no breaking-change handling, so `fix(providers)!` would have rendered under **Bug Fixes** as its description alone, with no BREAKING CHANGE section and no upgrade instruction — the entire justification for the major bump invisible to anyone reading the release rather than the commit. Confirmed historical: the `8.0.0` entry contains nothing but "release v8.0.0".

Fixed in `620ae1f`, verified by running git-cliff 2.13.1 against this branch rather than by reading the template — that file sits in the release critical path, and `publish-npm` declares `needs: release`, so a template that fails to parse would block the publish. Renders as:

```
### ⚠ BREAKING CHANGES

- OPENCODE_API_KEY no longer authenticates `zgo@` / `zengo@`. *(providers)*
```

### Accepted, not fixed here — LOW

Three findings recorded rather than fixed, to keep this release's blast radius assessable:

1. `providers/routing-hints.ts:62,74` — a user on the **bare-name** path (`kimi-k3` with only `OPENCODE_API_KEY`) gets `Set: export OPENCODE_GO_API_KEY=…` without the "is a DIFFERENT plan's key" clause. The key named is correct and actionable; only the explanation is missing, and only on that path.
2. `.github/workflows/release.yml:78` — the `sed … packages/cli/src/cli.ts` version injection matches nothing, since `VERSION` now lives in `version.ts`. Pre-existing silent no-op, harmless here only because `version.ts` was bumped correctly. Worth a separate fix.
3. `packages/cli/scripts/smoke/providers.ts:105-112` — still substitutes `"public"` for both OpenCode providers on the free-tier premise this release documents as measured-dead. Dev-only.

### Out of scope

The five Bun 1.4.0 failures are documented in `ai-docs/reports/bun-140-test-breakage-20260902.md`; the reviewer confirmed that write-up is correct. CI pins 1.3.10 and is green.

PR #235, opened 2026-09-07
chore: release v9.0.5

erudenko:
### Bug Fixes

- **mcp:** stop telling the agent to run `preflight` before `team`, `create_session` or `run_prompt`. Four directives in tool descriptions and result strings turned a diagnostic into a standing instruction; measured from transcripts, agents were spending up to four sequential 20-second probes per roster before doing any work. `preflight` stays, as a diagnostic for a human investigating a roster. A regression test guards all four sites.
- **diagnostics:** `preflight` and the TUI route probe no longer report bare Claude names (`claude-opus-5`, `opus`, `internal`) as "no route" or "OpenRouter, metered" and no longer advise dropping them. They are served natively on Claude Code's own auth, which these processes cannot probe; both now say so as a distinct "native (not probed)" outcome instead of folding it into success or failure. Slash-qualified ids (`anthropic/…`) still route to OpenRouter, matching the proxy.
- **team:** a slot cancelled before it drains stdin no longer raises an unhandled `EPIPE` stream error in the orchestrator (seen on macOS).

### Tests

- `route()` tests that assumed the Codex plan serves bare `gpt-5` now run against an empty catalog and skip when a real Codex login is present — Codex itself rejects `gpt-5` on a ChatGPT account, and the routing drop they tripped on is correct.

### Documentation

- CLAUDE.md gains the native-name invariant.
- Reports on the plan validation behind the `preflight` change, and on the `grok-4.6` Zen Go misroute — a stale 9.0.2 process after an auto-update, not a live defect.

### Verification

- Full suite on `ac712f0`: 3096 pass / 17 skip / 2 fail. The 2 are the `displayWidth` oracle tests, which disagree only under local Bun 1.4.0 vs CI's pinned 1.3.10; they pass in CI.
- Typecheck, lint (0 errors) and build clean.
- `preflight` behaviour proven on the built bundle over MCP stdio; the native fix went through four external review rounds, each REJECT verified and fixed.


erudenko:
## Code Review: claudish v9.0.5 release diff (origin/main...HEAD, 7 commits)

**Verdict**: PASS

**Summary**: The native-route helper faithfully mirrors the proxy's `isNative` condition (no `/` and no `@` in the target, then `parseModelSpec(...).provider === "native-anthropic"`), the parser tables it reads are a module-load snapshot of `BUILTIN_PROVIDERS` so calling it before and after `ensureEndpointsRegistered()` gives the same answer, the new `unverified` outcome is handled at every consumer (typecheck green), and the EPIPE handler is attached before the write. Typecheck passes and all four touched test files pass (84/84). The one substantive gap is that a typo'd bare name is now reported by `preflight` as a healthy native route with no negative signal at all, which the helper's own JSDoc says must not happen. No security findings.

### CRITICAL Issues (0)

### HIGH Issues (0)

### MEDIUM Issues (1)

#### Issue 1: `preflight` reports an unknown bare name as a healthy native route with no warning
- **Location**: `/Users/jack/mag/claudish/.claude/worktrees/team-fix2/packages/cli/src/mcp-server.ts:1013-1028`; contract stated in `/Users/jack/mag/claudish/.claude/worktrees/team-fix2/packages/cli/src/providers/native-route.ts:32-35`
- **Problem**: `parseModelSpec` sends every bare name that matches no provider pattern to `native-anthropic` (`model-parser.ts:259-266`), so `nativeRouteFor("claude-opus-5-tubro")` and `nativeRouteFor("some-typo-model")` return a route (the new test at `native-route.test.ts:44-52` pins this). `preflight` then emits a row with billing "native" and status "not probed — served on Claude Code's own auth", and counts it in neither `readyModels` nor `failedModels`, so the summary reads `Failed: 0` and the "Drop or replace before running" line never fires.
- **Why problematic**: `native-route.ts:32-35` says "Everything else — including a typo, which the proxy really does route natively — must be probed, never declared live", and exposes `isTierAlias` for exactly that split. Nothing consumes `isTierAlias` (zero callers outside the module) and both callers skip every native row unconditionally, so the documented contract and the shipped behaviour disagree. Before this diff the same input at least reached `route()` and failed; now the only trace of the typo is the wire-id column echoing the misspelling.
- **Impact**: An agent that asks `preflight({models: ["claude-opus-5-tubro", "glm-5.2"]})` sees one ready, zero failed, one native, and proceeds; the slot then dies at spawn with Claude Code's `unrecognized_model`, which is the minutes-into-the-run failure the tool description still promises to catch. The MEDIUM rating reflects that the tool is now explicitly a diagnostic rather than a precondition, so the wrong answer costs a slot, not a run.
- **Suggestion**: Distinguish "native and plausibly a Claude id" from "native only by fallthrough". The cheapest honest signal is a fourth row state for names that are neither a tier alias nor `claude-*` (e.g. "native by default — not a known Claude id, cannot verify") and counting those in the drop-or-replace line. If you keep `isTierAlias`, give it a consumer or drop it and rewrite the JSDoc so the helper does not promise a probe that no caller performs.

### LOW Issues (3)

#### Issue 2: Native-row reason strings are computed but never rendered
- **Location**: `/Users/jack/mag/claudish/.claude/worktrees/team-fix2/packages/cli/src/tui/hooks/useRouteProbe.ts:10-11`, `:153-154`, `:177-179`; consumer at `/Users/jack/mag/claudish/.claude/worktrees/team-fix2/packages/cli/src/tui/components/RoutingContent.tsx:274`
- **Problem**: The hook writes `reason: "Native Claude Code auth — served by the proxy, never routed"` on the initial row and immediately overwrites it with `REASON_NATIVE` in the same synchronous tick. `RoutingContent` never reads `entry.reason`; the label column is `PROVIDER_REASONS[entry.provider] ?? entry.provider`, and `PROVIDER_REASONS` has no `native-anthropic` key, so the panel row prints the literal `native-anthropic`.
- **Why problematic**: Two carefully worded explanations exist in state and neither is what the user sees; the next person to edit the copy will change the wrong string.
- **Impact**: Cosmetic; the badge and "Result:" line are correct.
- **Suggestion**: Add a `native-anthropic` entry to `PROVIDER_REASONS` and delete one of the two reason literals.

#### Issue 3: `const` declared between import statements
- **Location**: `/Users/jack/mag/claudish/.claude/worktrees/team-fix2/packages/cli/src/tui/hooks/useRouteProbe.ts:5-11`
- **Problem**: `REASON_NATIVE` sits between `import { nativeRouteFor }` and `import { describeProbeState }`. Biome accepts it (lint reports nothing for this file), but no other file in the package interleaves declarations with imports.
- **Impact**: Style only.
- **Suggestion**: Move the constant below the import block.

#### Issue 4: Two routing tests silently pass on any machine with a Codex login
- **Location**: `/Users/jack/mag/claudish/.claude/worktrees/team-fix2/packages/cli/src/providers/routing-rules.test.ts:733-734`, `:962-963`
- **Problem**: `if (existsSync(codexOauth)) return;` makes the test a no-op that still counts as a pass. The change is described as hermetic, but on a developer box with `~/.claudish/codex-oauth.json` the assertions never execute and nothing in the run output says so. The `cachePath` fix is real and works where the tests do run; this note is about the early return, which copies the sibling test at line 713.
- **Impact**: A regression in `openai-codex` rule composition would be invisible locally and only caught in CI.
- **Suggestion**: Express the condition with `test.skipIf(existsSync(codexOauth))` so the run reports a skip rather than a pass.

### Positive Observations
- `nativeRouteFor` checks `includes("/")` before parsing, which is the only way to reject `anthropic/claude-opus-5` (native-anthropic's own `/^anthropic\//` pattern would otherwise claim it); the test pins that case.
- `needsProxy` and the native branch call the same function against the same static table, so the "proxy is null but the loop wants to probe" state cannot arise; and an all-native roster no longer blocks on, or throws from, proxy startup.
- In `useRouteProbe` the native settle runs before `ensureProbeProxy()`, so a proxy-startup failure cannot repaint the native row red; `deriveProbeOutcome` is pure and its precedence (running > success > unverified > no-route) is fully tabled in the test.
- The `stdin` error listener is attached before `write()`, which is the ordering that actually matters for EPIPE; `channel/session-manager.ts:950` already had the equivalent, so the two spawn paths now agree.
- The preflight-directive test joins string-concatenation seams and self-checks that the join still works, so a broken regex fails loudly instead of making the guards vacuous.
- The new CLAUDE.md invariant names the trigger (`route()` on a bare Claude name), the cause (no credential store), and the third outcome, which is what the file's other entries do.
- Version bump is consistent across `package.json`, `packages/cli/package.json` and `packages/cli/src/version.ts` (all 9.0.5).

### Notes outside the diff
- `bun run lint` exits 1 with 40 biome warnings (all `noExplicitAny` and similar in `src/adapters/api-format.ts` and other untouched files). None of the flagged files appear in this diff, so it is pre-existing and not a blocker for this release, but a red lint gate is worth knowing about before tagging.

### Verdict Details
- **CRITICAL**: 0
- **HIGH**: 0
- **MEDIUM**: 1
- **LOW**: 3
- **Result**: PASS — no security findings, no wrong-behaviour-under-normal-input findings; the one MEDIUM is an honest-diagnostic gap on typo'd names that the code's own comments already identify. Safe to release; track Issue 1.

PR #249, opened 2026-09-13
fix(providers): send x-opencode-session so OpenCode Go/Zen requests route

oudivad:
# fix(providers): send `x-opencode-session` so OpenCode Go/Zen requests route

## Problem

Every model on the OpenCode Go tier fails in Claudish — `zengo@glm-5.3-flash`,
`zengo@deepseek-v4.1-flash`, all of them — with a valid key:

```
claudish --model zengo@deepseek-v4.1-flash "say hello"
# → 400 {"type":"error","error":{"type":"MissingSessionID",
#      "message":"Error from provider (Console Go): Request is missing
#                 x-opencode-session and cannot be routed efficiently.
#                 Please see https://opencode.ai/docs/go/#where-can-i-use-it"}}
```

The failure reads like a credential or model-gating problem, so it sends users
to check a key that is fine.

## Root cause

OpenCode Go requires a session id on every request. From the relay's own docs
("Where can I use it"), a client should:

> send a stable session ID in `x-opencode-session` for each conversation so we
> can optimize routing and prompt caching

and

> Identify itself with its own user agent … rather than a generic SDK or
> HTTP-library name.

`OpenAIProviderTransport.getHeaders()` emitted auth plus the provider's declared
`headers` and nothing else, and the `opencode-zen-go` definition declares no
`headers` — so the header was never sent. A grep for `x-opencode-session` across
the repository returns no hits.

Note that `model-discovery.ts` already carries a dated note about this exact
provider needing a `User-Agent` (a UA-less roster request returns Cloudflare
`403 error code: 1010`). The chat path sits behind the same edge and never got
the equivalent treatment.

## Change

`packages/cli/src/providers/transport/openai.ts`

- emits `x-opencode-session: claudish-<uuid>` for `opencode-zen-go` and
  `opencode-zen`;
- emits `User-Agent: claudish/<VERSION>` alongside it;
- both are written **before** the provider `headers` merge, so a definition that
  pins its own value still wins — the same precedence `model-discovery.ts` uses
  for its roster request.

The id is generated once per transport, which is the life of the conversation it
serves. That matters in both directions: a retry re-sends the **same** id, so a
turn cannot be re-routed mid-flight, while two concurrent conversations hold two
transports and get two ids, keeping their routing independent.

Nothing changes for any other provider: the block is gated on the provider name
and the emitted header set is otherwise identical.

## Verification

**Measured 2026-09-12**, one `POST /v1/chat/completions` per row, same key, same
body, the header the only difference:

| model | without `x-opencode-session` | with it |
|---|---|---|
| `glm-5.3-flash` | `400 MissingSessionID` | `200` |
| `deepseek-v4.1-flash` | `400 MissingSessionID` | `200` |

**Against the patched transport source** (the real `OpenAIProviderTransport`
compiled from the changed file, not a reimplementation), all checks pass:

```
PASS  x-opencode-session present — "claudish-519513ff-78bd-42b9-8a9d-a1dc76f0a21f"
PASS  id shape claudish-<uuid>
PASS  id stable across calls
PASS  User-Agent identifies the client — "claudish/9.3.0"
PASS  bearer auth still emitted
PASS  distinct id per conversation
PASS  header absent for other providers — ["Authorization"]
PASS  provider-declared value wins — "pinned"
PASS  live relay answers 200 — 200          (deepseek-v4.1-flash)
PASS  live body is a completion, not MissingSessionID
PASS  control (header removed) reproduces the 400 — 400
```

**Unit coverage added** in `openai.test.ts`: the header and its shape, its
stability across calls, its uniqueness per conversation, its absence for other
providers, and the definition-override precedence.

## Follow-ups (out of scope here)

- `opencode-zen` (the metered sibling) is included on the strength of sharing
  the apiPath, transport and relay; it has not been measured separately the way
  Go has.
- The Anthropic-format custom-endpoint path (`anthropic-compat` transport) would
  need the same header if someone points it at OpenCode Go.

### DIFF
--- a/packages/cli/src/providers/transport/openai.test.ts
+++ b/packages/cli/src/providers/transport/openai.test.ts
@@ -182,3 +182,82 @@ describe("isTerminal429", () => {
     expect(isTerminal429(transient)).toBe(false);
   });
 });
+
+/**
+ * OpenCode Go rejects any request that carries no `x-opencode-session`:
+ *
+ *   400 {"type":"error","error":{"type":"MissingSessionID","message":
+ *        "Error from provider (Console Go): Request is missing
+ *         x-opencode-session and cannot be routed efficiently."}}
+ *
+ * Measured 2026-09-12 on `/v1/chat/completions`: `glm-5.3-flash` and
+ * `deepseek-v4.1-flash` both return 200 with the header and this 400 without
+ * it, so the header is the whole difference. Before the fix, every Go model
+ * failed in Claudish with `zengo@…` despite a valid key.
+ */
+describe("OpenAIProviderTransport OpenCode session header", () => {
+  const goProvider: RemoteProvider = {
+    name: "opencode-zen-go",
+    baseUrl: "https://opencode.ai/zen/go",
+    apiPath: "/v1/chat/completions",
+    apiKeyEnvVar: "OPENCODE_GO_API_KEY",
+    prefixes: ["zengo@"],
+  };
+
+  test("opencode-zen-go carries an x-opencode-session id and identifies itself", async () => {
+    const headers = await new OpenAIProviderTransport(
+      goProvider,
+      "deepseek-v4.1-flash",
+      "test-key"
+    ).getHeaders();
+
+    // A `claudish-<uuid>` shape, not an empty placeholder: the relay routes on
+    // the value, so a constant or empty string would collapse every user onto
+    // one routing bucket.
+    expect(headers["x-opencode-session"]).toMatch(/^claudish-[0-9a-f-]{36}$/);
+    // The relay asks clients to name themselves rather than present a generic
+    // HTTP-library User-Agent.
+    expect(headers["User-Agent"]).toMatch(/^claudish\//);
+  });
+
+  test("the id is stable across calls, so a retry cannot move routing mid-turn", async () => {
+    const transport = new OpenAIProviderTransport(goProvider, "deepseek-v4.1-flash", "test-key");
+
+    const first = await transport.getHeaders();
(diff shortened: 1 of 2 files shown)


oudivad:
@claude review

erudenko:
Sorry for the slow reply on this one.

You were right about both headers, and the User-Agent half I'd missed entirely. That note in `model-discovery.ts` about a roster request getting Cloudflare `403 error code: 1010` without a UA has been sitting there since August, and it never occurred to me the chat path was behind the same edge. Good catch.

Both shipped in v9.4.0. I ended up writing it as a separate `OpenCodeZenTransport` subclass instead of a provider-name branch inside `openai.ts`, mostly because `getHeaders()` had no request parameter at all, so adding one meant touching the Layer 3 contract anyway.

The one real difference is where the session id comes from. Yours mints a uuid once per transport, and your reasoning holds for a normal CLI run: a retry re-sends the same id, two conversations get two transports. But transports are cached one per model, and `claudish serve` runs several conversations in a single process, so that uuid would quietly carry the first conversation's identity into the next one. v9.4.0 derives it per call from the inbound `metadata.user_id` instead, hashed, which is the same value Codex already sends as `prompt_cache_key`.

Closing this because the fix is live, not because anything in it was wrong. You found a bug that had been falling through to a per-token provider for anyone running a routing chain, and the commit says where the User-Agent came from. Thanks for putting the time in.

oudivad:
Thanks for the feedback, happy to help!

Le mer. 16 sept. 2026, 18:22, Jack Rudenko ***@***.***> a
écrit :

PR #250, opened 2026-09-14
fix(team,zen): v9.3.1 — Zen Go session header, team liveness for exited slots

erudenko:
### Fixed

- **OpenCode Zen Go works again.** Since 2026-09-12 Zen Go rejected every request
  with `400 MissingSessionID` ("Request is missing x-opencode-session"), so every
  `zengo@` / `zgo@` model failed, and inside a routing chain the request fell
  through silently to the next provider, often billed per token. claudish now
  sends `x-opencode-session` on every Zen request, including both retry paths.
  The value is a hash of Claude Code's own session id, the same key Codex already
  uses as `prompt_cache_key`. Verified live: `zengo@minimax-m3` returned 400 on
  9.3.0 and 200 on this build.
- **`team` status no longer reports finished slots as waiting for input.** While
  one slot was still running, a slot that had already exited kept appearing in
  `activity_by_slot` as `waiting_for_input`, with an idle count that grew from
  the moment it exited. The liveness maps now list running slots only, and a
  finished slot's reducer is settled to its real outcome.

### Documentation

- `ai-docs/architecture/adapters.md`: why a header that names the conversation is
  derived from the request on every call and never stored on a transport.
- `ai-docs/architecture/team-lifecycle.md`: why the liveness maps are RUNNING-only,
  and why `team` must settle the stream-json reducer itself.

### Verification

- Three new regression tests; each failed before the fix for the intended reason,
  and five mutation proofs turned them red again.
- Full hermetic suite, typecheck and lint: see the release gate results below.
- Root cause reviewed 3-0-0 and the patch 2-0-1 (both STRONG) by an internal
  reviewer plus external models.

### Known follow-ups

- OpenCode's docs also ask clients to send their own `User-Agent`; this release
  does not. Community PR #249 sends `User-Agent: claudish/<version>` and found
  the session header independently.
- The metered `opencode-zen` tier also sends the header, but is not verified live.


erudenko:
### Review of the delta on top of the reviewed fix

`dev:reviewer` judged the two commits added after `d675536` (the DeepSeek test hermeticity fix and the version bump). **Verdict: PASS** — 0 critical, 0 high, 3 medium, 2 low. Nothing blocks the release; the medium findings are test quality and become follow-ups.

**Medium**

1. `effort-mapping.test.ts:757` — the new test's second assertion cannot fail. `dsPrep` passes `thinking` only as the *original* request, so `request.thinking` is `undefined` whether the catalog opinion is honoured or not. The first assertion (`reasoning_effort`) is the one that discriminates, and the mutation check confirmed it: seeding `{ supported: true, control: "effort" }` failed with `Received: "high"`.
2. `seedDefaultCatalog` writes the real `~/.claudish/all-models.json` and restores it in `afterEach`. `scripts/guard-real-config.ts` snapshots only `config.json`, so a killed run can leave a synthetic catalog behind. Pre-existing pattern (used by the Sakana Fugu describe since 1325256); this change widens it from one describe to two.
3. Two retained titles ("deepseek-chat / deepseek-reasoner are V4 aliases", and the `deepseek-chat` half of "V4: none/minimal → thinking disabled") now pin the cold-cache name rule, while the live catalog makes shipped behaviour for `deepseek-chat` the opposite. Renaming them would stop a future reader "fixing" the dialect to match the title.

**Low**

4. `bun.lock` still records the `packages/cli` workspace version as 7.60.0; tolerated by `--frozen-lockfile` across every release since.
5. The Anthropic-wire branch for a `supported: false` DeepSeek id is not pinned (only the native wire is).

**Also verified by the review**: nested seeding restores strict LIFO on both the pass and fail paths, no sibling describe sees the synthetic catalog, no prior assertion was removed or weakened, and the bump is consistent across all three files with `version.ts` matching its generator.

Full report: `ai-docs/sessions/dev-fix-20260912-213141-f1fb0c1c/delta-review.md` (session directory, not committed).

PR #252, opened 2026-09-15
feat: add GitHub Models provider (gh@)

MayCXC:
Adds GitHub Models as a built-in `gh@` provider, the piece you said you wanted on its own in #81.

A single `BUILTIN_PROVIDERS` entry (`transport: "openai"`, base `https://models.github.ai`, path `/inference/chat/completions`, `shortcuts: ["gh"]`) reusing the shared `openaiHandler`. In the current merged-table design that `createHandler` is the wiring, so there is no separate profile line to forget and the "definition without a profile silently falls through to OpenRouter" hazard you flagged cannot happen here; the structural test `every builtin explicitly defines how its handler is created` covers it.

Prefix-only: no `nativeModelPatterns`, so a bare `gpt-*`/`o1-*` still routes to openai/openrouter and you opt in with `gh@`. Keyed by a fine-grained PAT with `models:read` (`GITHUB_MODELS_TOKEN`).

Tests: 4 added in `provider-definitions.test.ts` (first-class openai-transport built-in; shortcut and prefix resolve; key info and display name derive; prefix-only has no native pattern). Full suite and typecheck clean. Touches only `packages/cli/src`.

One thing to confirm before merge, as you noted in #81: the endpoint and PAT scope were months old then, so it is worth a live curl against `models.github.ai/inference/chat/completions` with a `models:read` PAT. I have not run that here (no token on this machine).


MayCXC:
Closing this: GitHub Models was retired for all customers on 2026-07-30 (https://github.blog/changelog/2026-07-30-github-models-is-now-retired/), and models.github.ai now answers a plain-text `OK` on every path, so there is no service left for this provider to route to.

PR #260, opened 2026-09-17
fix(routing): user routing overrides beat the native-anthropic heuristic

jsboige:
## Summary
- A bare name no prefix heuristic recognizes (e.g. a user-defined alias like `local-coding`) is classified `native-anthropic` by `parseModelSpec`, which made the proxy's fallback-chain block skip the routing engine entirely: the user's `routing` entry was never consulted and the request fell through to the Anthropic passthrough. `NativeHandler`'s proxyKey→Anthropic auth swap then strips the cluster key and the caller gets a misleading `401 "x-api-key header is required"` — even when the header IS present.
- Measured on the hub (fleet issue roo-extensions #3401): the `local-coding`/`local-fast` aliases were announced in `/v1/models` but every POST 401'd, while the same endpoint served the aliased model fine under its real name `qwen3.6-35b-a3b` (provider prefix `qwen` recognized → routing engine entered → user override applied → 200). Container logs confirm `model=local-coding handler=NativeHandler`.
- Fix: the fallback-chain guard now lets an exact or glob **user-authored** routing entry (excluding the `"*"` catch-all, which would divert every `claude-*` passthrough too) enter the routing engine, where the override wins as designed. Defaults ship globs only, so native passthrough behavior is unchanged for non-overridden names — verified against the live hub config: `local-coding`/`local-fast` → override, `claude-opus-4-8` → passthrough preserved.

## Test plan
- [x] 7 new tests in `routing-rules.test.ts` (`matchUserRoutingOverride`: exact, glob, catch-all exclusion, empty rules, GLM slug tolerance, end-to-end `route()` semantics for a native-classified alias)
- [x] `bun test packages/cli/src/providers/routing-rules.test.ts packages/cli/src/fork/` → 250/250
- [x] `bun run build` green
- [x] Live-config proof: `matchUserRoutingOverride` against the hub's real `~/.claudish/config.json` → both aliases true, `claude-opus-4-8` false
- [ ] Post-deploy observable: POST `local-coding` on the hub returns a real completion (blocked on the ACKed image-bump window 18/09 02:06Z)


jsboige:
Closing — opened against the wrong remote by mistake (this branch targets our internal fork's main, not this upstream). Apologies for the noise.

